PR #1, opened 2021-09-12
Rescue

pkaig:
* Rescue collective boost (high speed to gain altitude)
* After time delay drop to normal rescue collective speed (rescue_collective)
* If heli is inverted it will now roll to upright

Added CLI:
rescue_delay. (0-35) time delay in s *10. Set to 35 will disable roll from inverted
rescue_boost. Value added to rescue_collective for rapid collective until delay expires


pkaig:
I have updated to timer. Much nicer and reads better. Will close this request, update and resubmit.

PR #3, opened 2021-09-20
Docker files.

robhaswell:
Add simple Docker build container. This will allow users on systems with Docker installed to check out the repo and build the firmware with:

```bash
docker compose run rotorflight make arm_sdk_install
docker compose run rotorflight make STM32F411
```

After this is merged I can update the wiki.


rotorflight:
Integrated.

PR #5, opened 2021-11-29
MSP_COPY_PROFILE: re-initialize PID profile if needed

egonl:
The Lua scripts now support copying profiles. Unlike the Configurator you can also copy a profile to the current profile. In this case, pidInit should be called.


rotorflight:
Integrated with other stuff

PR #11, opened 2022-01-24
Move GS_LOST_THROTTLE state to isSpooledUp() == false

James-T1:
Throttle < min_check  (throttle hold) should instantly stop power to both motors.  Currently motorized tail continues to run for a few seconds if governor was previously active.  Impact of changing spooled up status will also begin iterm decay and prevent accumulation of absolute control error for this state.  Also, during a momentary loss of throttle signal such as an inadvertent flip of the throttle hold switch, the heli will start to spin when the tail motor stops (although not much since the main motor will also be at zero power).

### DIFF
--- a/src/main/flight/governor.c
+++ b/src/main/flight/governor.c
@@ -236,12 +236,12 @@ bool isSpooledUp(void)
         {
             case GS_ACTIVE:
             case GS_RECOVERY:
-            case GS_LOST_THROTTLE:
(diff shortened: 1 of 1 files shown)


rotorflight:
After considering this back and forth, decision was to leave it as is.
Anybody wanting the motors to stop immediately can set the lost throttle timeout to 0, which will disable the LOST_THROTTLE state and thus stop the motors straight away.

PR #14, opened 2022-04-16
default rates change to zero max rate

P-I-Engineer:
max rate....srates... give you extreme rate at end of stick travel akin to very thick expo. I do not believe this a good thing to have for LOS helicopter flying.
i've changed the defaults to be more linear and similar to Brain FBL.

current
image

proposed change
image


P-I-Engineer:
nevermind, this is not correct

PR #35, opened 2023-02-19
feat: implemented scale lights

egonl:
New LED parameters: blink pattern, blink pause, alternate color.
New LED overlays: Flicker ('K') and Dimmed ('D')
New profile: STATUS_DIMMED


rotorflight (approved):
Excellent!

PR #38, opened 2023-04-01
fix: added rcCommand[COLLECTIVE] to blackbox logging

egonl:
Added collective by refactoring RF1 for RF2. Now you can see also your stick movements again in the Blackbox Explorer.

image

Note that Throttle still isn't working properly - as before. @rotorflight Maybe you can look into this?

Test log with rcCommand[4] (collective) added.  

[LOG221.TXT](https://github.com/rotorflight/rotorflight-firmware/files/11129645/LOG221.TXT)

### DIFF
--- a/src/main/blackbox/blackbox.c
+++ b/src/main/blackbox/blackbox.c
@@ -201,7 +201,8 @@ static const blackboxDeltaFieldDefinition_t blackboxMainFields[] =
     {"rcCommand",   0, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
     {"rcCommand",   1, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
     {"rcCommand",   2, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
-    {"rcCommand",   3, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
+    {"rcCommand",   3, UNSIGNED, .Ipredict = PREDICT(0),       .Iencode = ENCODING(UNSIGNED_VB),  .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16), CONDITION(COMMAND)},
+    {"rcCommand",   4, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(SIGNED_VB), CONDITION(COMMAND)},
 
     /* setpoint - define 4 fields like RC command */
     {"setpoint",    0, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(SETPOINT)},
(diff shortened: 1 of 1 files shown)


rotorflight (changes requested):
Something is wrong with this change.
Collective was already in the log. It's rcCommand[3].  Throttle is rcCommand[4].
It is not logged. I guess you would like to add the missing throttle command?

rotorflight:
https://github.com/rotorflight/rotorflight-firmware/blob/devel/src/main/fc/rc_controls.h#L34

src/main/blackbox/blackbox.c
```diff
@@ -201,7 +201,8 @@ static const blackboxDeltaFieldDefinition_t blackboxMainFields[] =
     {"rcCommand",   0, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
     {"rcCommand",   1, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
     {"rcCommand",   2, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
-    {"rcCommand",   3, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16),  CONDITION(COMMAND)},
+    {"rcCommand",   3, UNSIGNED, .Ipredict = PREDICT(0),       .Iencode = ENCODING(UNSIGNED_VB),  .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(TAG8_4S16), CONDITION(COMMAND)},
```
rotorflight, line 204: This line is incorrect. rcCommand[3] is collective, so it must be SIGNED

src/main/blackbox/blackbox.c
```diff
@@ -201,7 +201,8 @@ static const blackboxDeltaFieldDefinition_t blackboxMainFields[] =
...
+    {"rcCommand",   4, SIGNED,   .Ipredict = PREDICT(0),       .Iencode = ENCODING(SIGNED_VB),    .Ppredict = PREDICT(PREVIOUS),      .Pencode = ENCODING(SIGNED_VB), CONDITION(COMMAND)},
```
rotorflight, line 205: And this is throttle, so it should be UNSIGNED

src/main/blackbox/blackbox.c
```diff
@@ -641,7 +642,17 @@ static void writeIntraframe(void)
     blackboxWriteUnsignedVB(blackboxCurrent->time);
 
     if (testBlackboxCondition(CONDITION(COMMAND))) {
-        blackboxWriteSigned16VBArray(blackboxCurrent->command, 4);
+        // Write roll, pitch and yaw first:
+        blackboxWriteSigned16VBArray(blackboxCurrent->command, 3);
```
rotorflight, line 646: This is incorrect. Above it is defined that there are 4 fields of type TAG8_4S16. But here we are writing only 3.

src/main/blackbox/blackbox.c
```diff
@@ -641,7 +642,17 @@ static void writeIntraframe(void)
...
+
+        /*
+        * Write the throttle separately from the rest of the RC data as it's unsigned.
+        * Throttle lies in range [PWM_RANGE_MIN..PWM_RANGE_MAX]:
+        */
+        blackboxWriteUnsignedVB(blackboxCurrent->command[THROTTLE]);
```
rotorflight, line 652: This is incorrect, for the same reason as above.

src/main/blackbox/blackbox.c
```diff
@@ -773,6 +784,10 @@ static void writeInterframe(void)
     if (testBlackboxCondition(CONDITION(COMMAND))) {
         CALC_DELTAS(deltas, blackboxCurrent->command, blackboxPrev->command, 4);
         blackboxWriteTag8_4S16(deltas);
+
+        // Calculate collective delta
+        int32_t collectiveDelta = blackboxCurrent->command[COLLECTIVE] - blackboxPrev->command[COLLECTIVE];
```
rotorflight, line 789: COLLECTIVE is already handled by the code two line before

src/main/blackbox/blackbox.c
egonl: Is it? Isn't it throttle? See the blackbox code.

src/main/blackbox/blackbox.c
egonl: I think it's collective, see the blackbox explorer code.

PR #39, opened 2023-04-01
fix: collective and throttle blackbox logging

egonl:
Processed comments from #38. Works only with Blackbox Explorer and PR 5, see https://github.com/rotorflight/rotorflight-blackbox/pull/5.


egonl:
Forgot to include some changes, PR updated.

rotorflight (approved):
perfect!

PR #40, opened 2023-04-23
feat: led strip 'arm state' tweak

egonl:
The 'arm state' mode now only shows armed/disarmed colors if there's no arming lock. If there is an arming lock, the background color will be shown, which is usually black.


pkaig (approved):
Looks fine to me.

PR #41, opened 2023-05-23
Cherry-pick ELRS binding bugfix

TNorthover:
First time here so apologies for any bad protocol. This is just a copy of BetaFlight PR https://github.com/betaflight/betaflight/pull/11845, which fixes ELRS (SPI) binding after another cherry-pick we took (80fe0be).


rotorflight:
We don't really support the SPI receivers, but it does not hurt to keep the code up-to-date for people who want to play with it. Thanks.

rotorflight (approved):

PR #46, opened 2023-08-31
dyn notch tuning

rtlopez:
Dynamic notch adaptations for heli
1. reduce minimum processing rate to 1k
2. min and max tracked frequency moved to 30-500
3. increased number of filters to 8 for each axis
4. by default is disabled, if enabled consumes about 10% CPU on F405, to enable set dyn_notch_count > 0 in CLI, 5 is resonable minimum if not using rpm filter
5. q factor is 3 initially
6. analyzer uses sliding dft and computer 72 frequency bins, so every bin is about 7Hz wide (500/72 for 1k sample rage, correct me if I'm wrong)
7. to calculate peaks surrounding shoulders are taken into account, so resolution is bit better
8. all 8 debug streams are utilized for `DEBUG_FFT_FREQ` mode
9. on the image below you can se, that peaks are easy to identify (select about 0.2s of log to analysis)
10. BE AWARE!!!, I haven't flown this firmware yet

rtfl_bb_fft


rtlopez:
Update:
1. lower the minimum frequency to 10Hz
2. reorganize `DEBUG_FFT_FREQ`, now `debug[0]` is pre-dyn-notch gyro data (after lpf), I'll provide blackbox viewer improvements soon

I did first test on the ground, here is the log 
[LOG00009.tar.gz](https://github.com/rotorflight/rotorflight-firmware/files/12507997/LOG00009.tar.gz)
It was tested on 450 class heli with paramters
```
dyn_notch_min_hz = 10
dyn_notch_max_hz = 245
dyn_notch_q = 250
dyn_notch_count = 6
```

At low RPMs you can see that the filter changes positions frequently. Only above 2000 rpm it starts to hold firmly, but you can also see that at these speed vibrations have increased significantly. Nevertheless, in the whole range the gyro trace is quite clean.

Note that, paramter `dyn_notch_max_hz` is connected with internal decimator, so setting `dyn_notch_max_hz` to 250 or lower cause that dyncamic filter downsample to integer denominator of loop rate (500Hz for 2k loop rate), thanks to this we can gain better resolution.

I wonder what is the safe lower limit of the filter. What is your experience with the RPM filter? What is usual range of main rotor speed during autorotation?

rotorflight:
Sorry about the late reply.

I think 10Hz is too low. Usually we don't want any filters below 20 or even 25Hz
Probably 25Hz is a good low limit.
If anybody is flying a large heli, they can change it to a lower value.

245Hz is perfect for maximum, as the Gyro hardware filter is set to 250Hz, and so is the decimation filter in RF.

rotorflight:
Can you use the DEBUG() macro instead of DEBUG_SET() macro.
Even though they do the same, the new macro is an indication that it has been updated for RF, and also that eight values may be used.

rotorflight:
And set the default min to 25 and default max to 245

rtlopez:
Hi @rotorflight and another batch of changes
1. I have disabled sorting peaks by frequency, now they sorted by amplitude and in the same order appear in blackbox, this reduces switching notches to other peaks especially for dominants.
2. This also affects notch Q factor advancing, now higher peak will get wider notch
3. There is no any kind of smooting tracked frequency, this mechanism caused delays in tracking and some noise leaked to result
4. I applied some patches from Betaflight to sdft
5. CPU usage increases about 8% after enabling filter, event on F411
6. I attach next log of test flight, I used following settings: `min_hz: 25, max_hz: 245, q: 200, count: 6` 

[LOG00041.tar.gz](https://github.com/rotorflight/rotorflight-firmware/files/12641715/LOG00041.tar.gz)

I still have a few ideas for improvements, but I feel pretty confident with this version and think it's safe for others to try. But before that, I would like to decide what to do next.
1. This branch is based on snapshot 0822 and needs to be updated to latest snapshot, what method do you prefer, merge latest devel to this branch, rebase, or none yet?
2. Q factor advancing is now hardcoded to 0.2, but it can be easily exposed to CLI and MSP. I can add this if you find it useful
3. I left some commented code yet, but I can remove it if you prefer

Some general settings recommendations 
1. **min_hz**: should be lower than main rotor frequency, but no lower that 20Hz, which is eqivalent of 1200rpm (but I noticed that the filter stops at about 5 Hz higher than in settings)
2. **max_hz**: should be about 10-20% higher than max tail rotor frequency, but no higher than 250Hz if filter rate is 1k. If filter rate is 2k or more, then 330-500 Hz can be used, especially for smaller helis or that with motorized tail (note that higher range means less precision and performance and requires more CPU)
3. **q**: is notch filter factor, lower number means wider filter, same rules apply as RPM filter
4. **q_advance**: notches are ordered by noise amplitude and every next notch can be more narrow, if this value is 0.1 and q is 2, then every next peak will be filtered with q: [2, 2.1, 2.2, 2.3, 2.4, ...] with this we can reduce filtering delay
5. **count**: number of tracked noise frequency, absolute minimum is 2, 4-6 is good for start if used standalone, lower number can be used in combination with RPM filter. Dynamic notch is applied after lowpass filters and RPM filter, so it will try to filter noise that passes there.

rotorflight:
Looking good.
1. You can rebase and force-push if you like. But definitely when it is ready. The branch must be based on the latest devel when I am about to merge it.
2. Sounds like a good value. Let's keep it like that for now. Too many parameters is not nice.
3. Remove commented code when it's ready.

rtlopez:
Hi @rotorflight I've rebased to the latest devel and tested on the ground. Would you like me to update the configurator? I see most code is still there, primarily limits and messages needs to be updated. Cheers.

rotorflight:
Looking good now. Are you finished with all changes? If so I can merge.

If you want to do the Configurator too, please go ahead!

rtlopez:
Hi @rotorflight I finished at the moment, you can merge it. Configurator update in linked PR.

rotorflight (approved):

PR #51, opened 2023-10-23
Dyn notch - more consistent filter latency

rtlopez:
I did some profiling on dyn notch and small experiment 
the idea here is that `dynNotchProcess` do it's job in 4 steps, and last two steps can be merged into one, as they are pretty short, then we get more consistent timings

dyn_notch_original_6_notches

dyn_notch_improve_6_notches

dyn_notch_improve_3_notches

in the original version timings are (us): 17, 24, 14, 19
in the updated version there are (us): 17, 25, 24
if there are only 3 notches, we get 16-19 us
most time consuming step is detect_peaks

Pros
- 3 steps allows to update notch filters more frequently
- overall timing is more consistent in every loop iteration
Cons
- overral processing time is bit higher in every loop iteration

### DIFF
--- a/src/main/flight/dyn_notch_filter.c
+++ b/src/main/flight/dyn_notch_filter.c
@@ -95,7 +95,7 @@ typedef enum {
     STEP_WINDOW,
     STEP_DETECT_PEAKS,
     STEP_CALC_FREQUENCIES,
-    STEP_UPDATE_FILTERS,
+    //STEP_UPDATE_FILTERS,
     STEP_COUNT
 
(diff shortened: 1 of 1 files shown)


rotorflight:
I will test this with the other scheduling changes I am working on. Good timing on this!

rotorflight:
It turns out the execution time depends heavily on the parameters used. Sometimes the fourth step is short, but with many notches it can be really long too. Best to keep it separate, as the total execution time with 3+4 together can grow too large.

PR #60, opened 2023-11-29
Re-instate check for too large no of rx channels.

AndersHoglund:
Fixes the problem with buffer over-run when using Rx and Tx with large no of channels. I.e. SRXL2 32 channels.


AndersHoglund:
DrRudder has a more complete solution. Closing.

PR #66, opened 2023-12-05
MPU gyro sample rate fix

rtlopez:
MPU6500 and derivatives (MPU9250, ICM206xx) ignores sample rate divider if DLPF is set to 0 and triggers EXTI interrupt with 8k/s. There are two solutions for this issue:
1. ability to skip specified number of readings in interrupt handler to achive requested samping rate.
2. force 1k internal sampling rate

both solutions are available, so it is possible to experiment

in gyroSetSampleRate() function
```c
// first solution to get 1k with DLPF=0
gyroRateKHz = GYRO_RATE_8_kHz;
gyroEXTIDenom = 8;

// second solution to get 1k internal sampling rate
gyroRateKHz = GYRO_RATE_1_kHz;
```

MPU6000 is not affected

### DIFF
--- a/src/main/drivers/accgyro/accgyro.h
+++ b/src/main/drivers/accgyro/accgyro.h
@@ -112,6 +112,8 @@ typedef struct gyroDev_s {
     uint32_t detectedEXTI;
     uint32_t gyroLastEXTI;
     uint32_t gyroSyncEXTI;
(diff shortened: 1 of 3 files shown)


rotorflight:
Refactored and merged manually.

PR #67, opened 2023-12-07
Horizon Mode Inverted Self-Level

KenImhof:
Horizon mode modification for the helicopter to self-level when Inverted in addition to the typical upright self-level operation. This makes it easier for new pilots to lean inverted flight.

This needs testing as I could not get the DEVEL branch to run with my snapshot version of the configurator.

### DIFF
--- a/src/main/flight/imu.c
+++ b/src/main/flight/imu.c
@@ -634,3 +634,9 @@ bool isUpright(void)
     return true;
 #endif
 }
+
+bool isUpsidedown(void)
+{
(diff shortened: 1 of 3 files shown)


AndersHoglund:
Please squash those three commits into one. A PR should only have one commit, containing all changed files and dependencies in one atomic update. With a proper description as well. In this case the PR title for example, not only "Update file.c". git can provide info on which files are changed, but not why.

rotorflight:
No activity.

PR #77, opened 2024-02-15
YGE ESC telemetry sensor implementing OpenYGE

bob01:
New protocol developed by YGE for open source projects like RotorFlight (Thanks F.F.)

This sensor is full featured, providing all sensor values supported by the existing RotorFlight telemetry framework
- temp, voltage, current, consumption, rpm, throttle pwm and bec temp (as temperature 2)

Several other  values like bec_voltage, bec_current, ESC status etc. are available with RotorFlight telemetry framework extensions found on the 'esc_sensor_data' branch which is a superset of this one.

* The telemetry signal from the ESC needs to be connected to the TX pin of the selected UART - the serial port is opened in bi-directional mode to facilitate upcoming ESC setup/programming functionality for supported ESCs.

* Matching PR which includes "OpenYGE" in the ESC telemetry sensor list on the configurator [ Motor ] is
https://github.com/rotorflight/rotorflight-configurator/pull/77

Many thanks to YGE for working with us to make this happen and in record time.
* Requires YGE ESC firmware v1.03543 or greater (available soon)


src/main/sensors/esc_sensor.c
```diff
@@ -1577,13 +1843,19 @@ bool INIT_CODE escSensorInit(void)
         case ESC_SENSOR_PROTO_APD:
             baudrate = 115200;
             break;
+        case ESC_SENSOR_PROTO_OPENYGE:
+            callback = oygeDataReceive;
+            baudrate = 115200;
+            mode = MODE_RXTX;
+            options |= SERIAL_BIDIR;
```
rotorflight, line 1850: What is this for?
It would imply the telemetry must be connected to the TX pin on the FC.
The other telemetry protocols are connected to the RX pin.

src/main/sensors/esc_sensor.c
```diff
@@ -1478,6 +1478,268 @@ static void apdSensorProcess(timeUs_t currentTimeUs)
...
+    int16_t  temp = buffer[5];
+    uint16_t volt = buffer[7] << 8 | buffer[6];
+    uint16_t curr = buffer[9] << 8 | buffer[8];
+    uint16_t capa = buffer[11] << 8 | buffer[10];
+    uint16_t erpm = buffer[13] << 8 | buffer[12];
+    uint8_t   pwm = buffer[14];
+    uint16_t voltBEC = buffer[17] << 8 | buffer[16];
+    int16_t  tempBEC = buffer[20];
+
+    if (version >= 2) {
```
rotorflight, line 1661: Would version 1 ever be released by YGE?

src/main/sensors/esc_sensor.c
bob01: Needed for bi-directional (single wire serial communication) between the FC and connected devices with a single signal wire connection - both devices are connected by TX pin only allowing the use of 2 lead (or 3 lead connectors if +ve is needed).

Bi-dir is required for ESC setup/programming (if supported by the ESC).
Coming soon for YGE - I'd like to get Scorpion and OMP done too
image

Doing it now appears to have no downsides vs later which would require telemetry cablers in use to be modified (or resources remapped)

src/main/sensors/esc_sensor.c
bob01: It might - the temp offset change may/may make it into YGE's firmware RC - we may be out of touch for the next 2 weeks. If YGE dev was able to get the change in it will be v2, if not then v1.

src/main/sensors/esc_sensor.c
rotorflight: The downside is that the release won't have pin swapping in the Configurator, or in the serial drivers.
Users would need to have a different configuration for YGE telemetry setup, likely with "complex" CLI commands.
As the Tx is not used right now, I suggest you keep it Rx only for the forthcoming release, so that it works the same as the other protocols.
The two-way development can happen after the release.

src/main/sensors/esc_sensor.c
rotorflight: that's ok

src/main/sensors/esc_sensor.c
bob01: Agreed, RX for initial release.
RX / TX matter will be much easier deal with if/when pin remapping comes to the configurator
Thanks

PR #84, opened 2024-03-02
Angle Mode setpoint input arg was lost. 

AndersHoglund:
Noticed different ele/ail stick feel around center with RF2-RC1. In angle mode very unresponsive, horizon mode more normal and like acro mode. In RF1 there was no such difference. Turns out the angleModeApply function overwrites the input arg pidSetpoint while horizonModeApply does not.  Angle and Horizon modes now handles setpoint arg the same way.

### DIFF
--- a/src/main/flight/leveling.c
+++ b/src/main/flight/leveling.c
@@ -174,7 +174,7 @@ float angleModeApply(int axis, float pidSetpoint)
         if (!isAirborne())
             errorAngle *= 0.25f;
 
-        pidSetpoint = errorAngle * level.Gain;
+        pidSetpoint += errorAngle * level.Gain;
     }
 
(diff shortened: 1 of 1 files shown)


rotorflight:
Very interesting find!
I will double-check where/when it happened.
Funny that nobody noticed.

AndersHoglund:
Did some more checking and the difference in Angle mode feel compared to RF1 is of course the 0.25 reduction when not airborne. Removing this condition for both Angle and Hor modes makes them feel much more similar around center. And makes RF2 feel like RF1.
I was fooled by Angle and Horizon feeling so different, why I still do not know. Probably the 0.25 factor is less significant in Hor mode.
Closing this PR as the suggested fix is wrong.

PR #87, opened 2024-03-15
OpenYGE - fix to support v3 frame header extension

bob01:
Fix required for OpenYGE telemetry  v3 frames which may appear with YGE FW release.
It was necessary to extend the header which will cause a compatibility issue unless addressed.

### DIFF
--- a/src/main/sensors/esc_sensor.c
+++ b/src/main/sensors/esc_sensor.c
@@ -1650,14 +1650,16 @@ static uint8_t oygeDecodeTelemetryFrame(void)
     }
 
     uint8_t version = buffer[1];
-    int16_t temp = buffer[5];
-    uint16_t volt = buffer[7] << 8 | buffer[6];
-    uint16_t curr = buffer[9] << 8 | buffer[8];
-    uint16_t capa = buffer[11] << 8 | buffer[10];
-    uint16_t erpm = buffer[13] << 8 | buffer[12];
-    uint8_t pwm = buffer[14];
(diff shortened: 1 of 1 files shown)


rotorflight:
All PRs into the master branch, please.

bob01:
Np - thought bug fixes went to release and things not meant for the pending releasee went to master - closing...

rotorflight:
Generally speaking, anything going into the release branches must be first validated in the master branch, usually by people testing a snapshot made from the master, or a test package. Once we are happy that everything is totally fine with the commit, then we can consider adding it to the release branch.

rotorflight:
Is the YGE FW release using v3 format, and some earlier test versions are using v1 and v2?
If so, is it worth keeping v1 and v2 around? Any reason for people to not update to the YGE release?

bob01:
There is a strong possibility at this point if the YGE FW release will be v3+ but uncertain.
The v3 data frame is significant because it is a header extension which breaks existing listener implementations (RF).
YGE FW snapshot 1.03547 (v2) is already in the public domain outside the test group (unexpected) - have to assume there will be a mix of v2 and v3+ plus around for some time.

bob01:
Testing of this fix was limited to small number of subjects (3 helis, 2 pilots) w/ YGE FW 1.03547 prior to submission of the PR.

rotorflight:
Okay... That's rather poor development practise from YGE.
No problem. We can keep them.

rotorflight (approved):

PR #93, opened 2024-03-26
Horizon inverted mode self-leveling

KenImhof:
Horizon mode inverted self-leveling rather than flipping back upright.

### DIFF
--- a/src/main/flight/imu.c
+++ b/src/main/flight/imu.c
@@ -637,9 +637,5 @@ bool isUpright(void)
 
 bool isUpsidedown(void)
 {
-#ifdef USE_ACC
-    return !sensors(SENSOR_ACC) || (attitudeIsEstablished && getCosTiltAngle() < 0);
-#else
-    return false;
-#endif
+    return sensors(SENSOR_ACC) && attitudeIsEstablished && getCosTiltAngle() < 0;
 }
--- a/src/main/flight/imu.c
+++ b/src/main/flight/imu.c
@@ -640,6 +640,6 @@ bool isUpsidedown(void)
 #ifdef USE_ACC
     return !sensors(SENSOR_ACC) || (attitudeIsEstablished && getCosTiltAngle() < 0);
 #else
-    return true;
+    return false;
 #endif
 }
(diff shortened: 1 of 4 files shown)


src/main/flight/leveling.c
```diff
@@ -100,8 +100,19 @@ static float calcLevelErrorAngle(int axis)
 #endif
     angle = constrainf(angle, -level.AngleLimit, level.AngleLimit);
 
-    float error = angle - ((attitude.raw[axis] - angleTrim->raw[axis]) / 10.0f);
-
+    float currentAngle = ((attitude.raw[axis] - angleTrim->raw[axis]) / 10.0f);
```
rotorflight, line 103: Please fix indentation. Four spaces. Dont use tabs.

src/main/flight/imu.c
```diff
@@ -634,3 +634,12 @@ bool isUpright(void)
     return true;
 #endif
 }
+
+bool isUpsidedown(void)
+{
+#ifdef USE_ACC
+    return !sensors(SENSOR_ACC) || (attitudeIsEstablished && getCosTiltAngle() < 0);
```
rotorflight, line 641: This does not look correct logically.
If ACC sensor is not in use, this will report true = upsidedown all the time.
I think it should report false in that case.

src/main/flight/imu.c
```diff
@@ -634,3 +634,12 @@ bool isUpright(void)
...
+#else
+    return true;
```
rotorflight, line 643: Same here. return false.

rotorflight (changes requested):

rotorflight:
Has this been tested by anybody in the beta-tests team in Discord?

KenImhof:
Yes, I am running that code and I can make a short video if that helps.

KenImhof:
Short Flight Test Video:
https://www.youtube.com/watch?v=_i-cUbpH7xk

egonl:
I've tested the code and I think it's good to go into RC2.

src/main/flight/imu.c
egonl: Returning false now

src/main/flight/imu.c
rotorflight, line 641: This says the heli is upside-down if ACC is not enabled. This can't be right?
Please fix.

rotorflight (changes requested):

src/main/flight/imu.c
rotorflight: Should be:
```
return sensors(SENSOR_ACC) && attitudeIsEstablished && getCosTiltAngle() < 0;
```

rotorflight:
Merged manually

PR #98, opened 2024-04-14
Fix for current FW Kontronik ESCs - verified Kolibri v3.0/v3.5, Kosmik 200 HV v4.17

bob01:
Current ESC FW for the Jive (v1.17) / Kolibri (3.0 - 3.5) / Kosmik (v4.17) are 40 byte data + 44 byte info.
JivePro/Kosmic v*.13 were different 38 byte w/ 32 byte payload, 2 extra bytes and 4 byte CRC, based on release notes it is likely that the *.14 versions switched to the current 40 byte packet but this could not be verified due to lack of test subjects.

This is a simple fix to support modern 40 byte packet - a more elaborate scheme to support older controllers is outside of RF RC scope, possible future consideration.

Kosmik 250 HV-I could not be properly verified due to what appears to be an unrelated defect - versions 1.17 - 1.25 appeared to produce correct 40 byte but broken frames.

### DIFF
--- a/src/main/sensors/esc_sensor.c
+++ b/src/main/sensors/esc_sensor.c
@@ -1008,7 +1008,8 @@ static void uncSensorProcess(timeUs_t currentTimeUs)
  *  28-31:      Error Flags
  *     32:      Operational condition
  *     33:      Timing 0..30
- *  34-37:      CRC32
+ *  34-35:      Reserved
+ *  36-39:      CRC32
  *
  */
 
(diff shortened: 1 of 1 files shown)


rotorflight (changes requested):
Can you also fix the protocol listing above. I think it doesn't list the two reserved bytes.

bob01:
Sure - this was also accidentally checked into master - fixing that - will likely resubmit PR

bob01:
Suggest reviewing https://github.com/rotorflight/rotorflight-firmware/pull/99 instead.

rotorflight (approved):

PR #99, opened 2024-04-14
Auto detection fix for Kontronik ESCs - verified Kolibri v3.0/v3.5, Kosmik 200 HV v4.17, JivePro 120 HV v1.13

bob01:
Enhancement to https://github.com/rotorflight/rotorflight-firmware/pull/98 to handle legacy and current Kontronik  ESCs

- assume length unknown
- while length unknown
- test at 38 bytes for valid CRC for 32 or 34 byte payload, set expected length == 38 (and crc exclusion) if yes
- test at 40 bytes for valid CRC, set expected length = 40 if yes

- use length determined for all subsequent packets

### DIFF
--- a/src/main/sensors/esc_sensor.c
+++ b/src/main/sensors/esc_sensor.c
@@ -1012,6 +1012,12 @@ static void uncSensorProcess(timeUs_t currentTimeUs)
  *  36-39:      CRC32
  *
  */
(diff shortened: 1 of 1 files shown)


bob01:
Replaced by - https://github.com/rotorflight/rotorflight-firmware/pull/100
Branch cleanup - no code changes,

PR #100, opened 2024-04-15
Auto detection fix for Kontronik ESCs - verified Kolibri v3.0/v3.5, Kosmik 200 HV v4.17, JivePro 120 HV v1.13

bob01:
Replaces https://github.com/rotorflight/rotorflight-firmware/pull/99 (branch cleanup, single commit cherry picked)

Enhancement to https://github.com/rotorflight/rotorflight-firmware/pull/98 to handle legacy and current Kontronik ESCs
- assume length unknown
- while length unknown
- test at 38 bytes for valid CRC for 32 or 34 byte payload, set expected length == 38 (and crc exclusion) if yes
- test at 40 bytes for valid CRC, set expected length = 40 if yes, if no as usual

- when length known
- use length determined for all subsequent packets

### DIFF
--- a/src/main/sensors/esc_sensor.c
+++ b/src/main/sensors/esc_sensor.c
@@ -1085,7 +1085,7 @@ static bool processKontronikTelemetryStream(uint8_t dataByte)
     }
     else if (kontronikPacketLength == 0 && readBytes == KON_FRAME_LENGTH) {
         // auto detect 40 byte packet...
-        uint32_t crc = kontronikDecodeCRC(40 - KON_CRC_LENGTH);
+        uint32_t crc = kontronikDecodeCRC(KON_FRAME_LENGTH - KON_CRC_LENGTH);
         if (calculateCRC32(buffer, KON_FRAME_LENGTH - KON_CRC_LENGTH) == crc) {
             // ...w/ 36 byte payload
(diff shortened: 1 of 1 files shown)


src/main/sensors/esc_sensor.c
```diff
@@ -1050,7 +1061,36 @@ static bool processKontronikTelemetryStream(uint8_t dataByte)
...
+            // ...w/ 34 byte payload
+            kontronikPacketLength = KON_FRAME_LENGTH_LEGACY;
+            kontronikCrcExclude = 0;
+            readBytes = 0;
+            return true;
+        }
+    }
+    else if (kontronikPacketLength == 0 && readBytes == KON_FRAME_LENGTH) {
+        // auto detect 40 byte packet...
+        uint32_t crc = kontronikDecodeCRC(40 - KON_CRC_LENGTH);
```
rotorflight, line 1088: Other frame sizes are #defines, 40 is not. Please fix.

rotorflight (changes requested):

src/main/sensors/esc_sensor.c
bob01: Ops - done. 
Verified w/ Kolibri v3.5.

Issue #101, opened 2024-04-18
ESC Setup / programming (bidirectional telemetry)


bob01:
Provide ESC setup / programming functionality using the Tx via bidirectional telemetry on par with other modern R/C systems.

As heli pilots we are accustomed to being able to setup (set BEC voltage, change rotation, set cutoff behavior etc.) and tweak at the field (ESC internal gov gains etc). A growing number of ESCs support this functionality.
Support in RF.

rotorflight:
Feel free to open PRs into master.
Release is on the RF- branch already.

bob01:
Will do.

rotorflight:
Work done for now.

Issue #78, opened 2024-02-17
Extend ESC telemetry data


rotorflight:
Some ESCs are providing more data than the FC can use at the moment.
The ESC telemetry support should be extented to handle this.

bob01:
Mind if I take this one on?

bob01:
On it.

rotorflight:
All done.

PR #108, opened 2024-04-28
ESC Setup / programming (bidirectional telemetry / forward programming) - implemented for YGE and Scorpion ESCs

bob01:
ESC setup / programming (forward programming) using the Tx via bidirectional telemetry #101, includes #78

Complete for YGE and Scorpion ESCs.
HobbyWing V5 in the works, POC looking good -available soon if no showstoppers.

- Both implemented on a common request / response finite state machine (rrfsm) which handles bidirectional and unidirectional communication with the ESC. Adding other ESC protocols is relatively easy.
- Common ESC parameter read(collection) and write pattern
- Common MSP transport w/ opaque payload serving all supported ESC between FC and Tx,
ie transport is common, serialization/deserialization of payload at FC is part of per ESC support code.
Transmitter tools are outside the scope of this project - per manufacturer scripts not tied to RF release cycle.
Working reference implementations for YGE, Scorpion and HobbyWing (soon) are provided.

### Configuration
- enable one-wire bidirectional (half-duplex) communication on ESC telemetry sensor port
set esc_sensor_halfduplex = ON
- half duplex operation requires the ESC to be connected to the Tx pin instead of the Rx pin
Instead of modifying our telemetry cables we opt to address this with resource configuration as some of our chosen flight controllers don't have Tx pins on some uarts and we don't want to create non-standard telemetry cables. 
e.g. for FDF722 v2
\# resources
resource SERIAL_RX 3 NONE
resource SERIAL_TX 3 C11

### Code
- escSensorInit: sets callback to function returned by protocol initialization method which shall initialize protocol and rrfsm as appropriate for half or full duplex operation. rrfsmDataReceive usually returned.
- escSensorProcess: delegates to rrfsmSensorProcess(...) which is a basic FSM which...
   - handles reporting of sync, crc errors, timeouts, frame counts
   - sequencing write activity
   - delegating configured handlers for accepting and decoding incoming character stream
   - handling self echo on writes
- no protocol specific code here.

- 2x new MSP commands
   - MSP_ESC_PARAMETERS (217)
   Writes paramBuffer if complete set of ESC parameters received from device and ready for editing, false otherwise
   - MSP_SET_ESC_PARAMETERS (218)
   Reads into paramUpdBuffer then calls paramCommit callback (configured in protocol init) which returns true on success or false otherwise
   
- Common parameter patterns
   - interface is paramBuffer, paramUpdBuffer, paramPayloadLength and paramCommit(...)
   - paramBuffer and paramUpdBuffer are exchanged by MSP as described above. There is a short header which contains
      - ESC specific signature: detects mis-matches between ESC and Tx app (Lua script for RadioMaster)
      - flags+version: param protocol version, read-only flag, user option (e.g. ESC reset)
   - protocol is considered to be in 'readonly' mode if paramCommit is not set (say if esc_sensor_halfduplex == OFF).
   UI shall not display 'Save' or 'User' options in this case
   - parameters collected / read from ESC in to paramBuffer
   - parameters considered 'ready' only when complete set available, protocol indicates ready by setting paramPayloadLength to the length of data in paramBuffer, 0 otherwise.
   UI shall display 'ESC not ready...' in this case
   - protocol's paramCommit implementation shall perform change detection between paramUpdBuffer received from MSP and paramBuffer to avoid unnecessary write operations against the ESC. Dirty parameters shall be indicated.
   
- Common RRFSM pattern
   - configurable boot delay as most ESCs take several seconds to fully initialize
   - assuming the master role
      - if ESC starts sending frames after init FC shall wait for first frame and send first request shortly after. This avoids the need for collision detection. Specific examples below.
      - if ESC starts in slave mode FC shall send first request during initial quiet time
   - next request is scheduled for frame-period after previous response. Pending requests are sent one at a time until there are none remaining in the following order...
      - special commands like ESC reset if supported / needed are first
      - pending ESC write requests to update parameters flagged dirty are scheduled next. Dirty flag is immediately reset - there are no retries.
      - pending ESC read requests to fetch stale or missing parameters are scheduled next
      - request next telemetry frame if nothing else

Each ESC's behaviour, capabilities and parameter set is different - discussed below
#### YGE support
- configure FC and ESC telemetry protocol as "OpenYGE"
- Requires OpenYGE v3 (FW 1.03551) or later
- initializes in AUTO_TELE mode and begins sending unsolicited TELE_AUTO packets (like full duplex operation)
- FC assumes master role after receipt of first TELE_AUTO packet and sends the first TELE_REQ packet
(unless is < v3 packet in which case FC reverts to AUTO_TELE mode and will not try to assume master role until next bootup)
- ESC assumes slave role, stops send unsolicited packets and responds with TELE_RESP packet
- FC schedules next TELE_REQ request on receiving response for next frame period. Exact timing is imprecise '38' was found to produce a result close enough to 50ms frame period
- each REQ packet header contains a rotating sequence number, matching response header will include this seq number
- YGE parameters are 16 bit indexed values. v3 defines 32
- ESC sends 1 parameter round-robin with each telemetry packet so complete collection is received in about 32 x 50ms in  1600ms. There is no need for an explicit 'get parameter(s)' request. Collection is made available to MSP when all have been received.
- oygeParamCommit(...) applies change detection to avoid unnecessary writes, sets each. changed parameter's dirty bit and invalidates it in the parameter collection, collection is now unavailable to MSP
- FC schedules WRITE_PARAM_REQ (instead of TELE_REQ) for dirty param and clears param's dirty bit - there are no retries
- ESC responds with WRITE_PARAM_RESP which is exactly the same format as regular TELE_RESP but will break the parameter round-robin and include the parameter value just written. Parameter is cached as before and if collection is complete it is once again made available to MSP
- User may view and change parameters at any time when safe to do so (on the ground/disamed recommended of course). 
Changes are immediate, no restart required - can be done as often as needed, e.g. setup, flight tuning internal gov PIDs etc.
Note: ESC will ignore write requests if the motor is turning
Note: All ESC responses are in the same frame format w/ TELE_AUTO being slightly different as it's seq # will always be 0.
Note: ESC sends parameters even in AUTO_TELE mode ie these can be viewed on the Tx (read only MSP) even if when serial port not configured for bidirectional communication

#### Scorpion Tribunus
- Requires v63 or later
- configure FC and ESC telemetry protocol as 'UNC'
initial req / response implementation as per documentation for 'Scorpion' protocol was functional but contents were encrypted so was set aside and later removed in favour of UNC workaround - may be revisit later
- ESC waits for ~ 10 seconds after power on, if a request to read status region is received during this time it will abort entry to UNC mode permanently / until reset
- if undisturbed ESC enters UNC mode, starts sending telemetry data and stops respond to requests, just like before
- if MSP activity is detected during this quiet time flag paramMspActive is set
- UNCSETUPFSM detects paramMspActive and sends a READ_STATUS request - response confirms that UNC was successfully stopped, UNCSETUPFSM into ACTIVE state
- UNCSETUPFSM triggers the next request pattern as needed. Once triggered the next request is triggered by the previous response until there are no more
   - commands first
   - parameter writes, next
   - parameter reads
   - back to UNCSETUPFSM
- paramBuffer is packed with ranges of parameters from multiple regions - these ranges are declared in tribParamAddrLen
- entries from tribParamAddrLen are used...
   - to build ESC read requests - each entry specifying region, address and length - results are packed into paramBuffer for MSP, parameters made available when collection is complete
   - perform change detection on paramUpdBuffer in paramCommit(...) and later build ESC write requests from changes unpacked from paramUpdBuffer from MSP
- User must be ready to enter any page in the Tx app ASAP to abort UNC to view and change parameters therefore this option is available only at startup.
Restart required to apply changes and restart UNC telemetry data
- 'ESC Reset' provided as param 'user option' and is presented along with 'Save' on menu
- There is a certain interesting thing that can be done because this setup communication happens outside of the ESC's configured protocol

#### Testing
- YGE: Tested w/ several 205HV, 135 Aureus v1 and v2's since mid-March - has been in general use in more than 6 helis for the last month or so.
- Scorpion: Tested w/ 80A and 130A ESCs (v63) in 2 helis since late Feb
No known outstanding issues with either.

#### UI for YGE
image
image
image
image
image
image
#### UI for Scorpion - connect to Scorpion ASAP on startup, reset when done (power cycle or use menu option)
image
image
image
image
image
image
image
image
image


src/main/sensors/esc_sensor.c
```diff
@@ -1106,21 +1008,30 @@ static void kontronikSensorProcess(timeUs_t currentTimeUs)
             uint32_t crc = kontronikDecodeCRC(kontronikPacketLength - KON_CRC_LENGTH);
             if (calculateCRC32(buffer, kontronikPacketLength - kontronikCrcExclude - KON_CRC_LENGTH) == crc) {
                 uint32_t rpm = buffer[7] << 24 | buffer[6] << 16 | buffer[5] << 8 | buffer[4];
+                int16_t  throttle = (int8_t)buffer[24];
```
rotorflight, line 1011: Can this be negative? Why the (int8_t)?

src/main/sensors/esc_sensor.c
```diff
@@ -1106,21 +1008,30 @@ static void kontronikSensorProcess(timeUs_t currentTimeUs)
...
                 uint16_t pwm = buffer[23] << 8 | buffer[22];
                 uint16_t voltage = buffer[9] << 8 | buffer[8];
                 uint16_t current = buffer[11] << 8 | buffer[10];
                 uint16_t capacity = buffer[17] << 8 | buffer[16];
                 int16_t  tempFET = (int8_t)buffer[26];
                 int16_t  tempBEC = (int8_t)buffer[27];
+                uint16_t currBEC = buffer[19] << 8 | buffer[18];
+                uint16_t voltBEC = buffer[21] << 8 | buffer[20];
+                uint16_t status1 = buffer[29] << 8 | buffer[28];
+                uint16_t status2 = buffer[31] << 8 | buffer[30];
```
rotorflight, line 1021: Why not collect the status directly into a uint32_t?

src/main/sensors/esc_sensor.c
```diff
@@ -1518,6 +1440,658 @@ static void apdSensorProcess(timeUs_t currentTimeUs)
...
+{
+    rrfsmStartFrame(currentTimeMs);
+
+    rrfsmFramePeriod = 0;
+    readIngoreBytes = reqLength;
+    if (reqLength != 0)
+        serialWriteBuf(escSensorPort, reqbuffer, reqLength);
+}
+
+static void rrfsmInvalidateReq()
```
rotorflight, line 1494: C-convention. Should be (void)

src/main/sensors/esc_sensor.c
```diff
@@ -1917,4 +2620,30 @@ bool INIT_CODE escSensorInit(void)
...
 }
 
+
+uint8_t escGetParamBufferLength(void)
+{
+    paramMspActive = true;
+    return paramPayloadLength != 0 ? PARAM_HEADER_SIZE + paramPayloadLength : 0;
+}
+
+uint8_t *escGetParamBuffer()
```
rotorflight, line 2630: (void)

src/main/sensors/esc_sensor.c
```diff
@@ -1917,4 +2620,30 @@ bool INIT_CODE escSensorInit(void)
...
+{
+    paramBuffer[PARAM_HEADER_SIG] = paramSig;
+    paramBuffer[PARAM_HEADER_VER] = paramVer | (paramCommit == NULL ? PARAM_HEADER_RDONLY : 0);
+    return paramBuffer;
+}
+
+uint8_t *escGetParamUpdBuffer()
```
rotorflight, line 2637: (void)

src/main/sensors/esc_sensor.c
```diff
@@ -1917,4 +2620,30 @@ bool INIT_CODE escSensorInit(void)
...
+{
+    return paramUpdBuffer;
+}
+
+bool escCommitParameters()
```
rotorflight, line 2642: (void)

src/main/sensors/esc_sensor.c
```diff
@@ -1518,6 +1440,658 @@ static void apdSensorProcess(timeUs_t currentTimeUs)
...
+        const uint8_t len = *pal & TRIB_PAL_LEN_MASK;
+        if ((dirtybits & 0x01) != 0) {
+            void *payload;
+            uint32_t iq22Payload[2];
+            if (addr == TRIB_PARAM_IQ22_ADDR) {
+                // convert scaled uint32 -> IQ22
+                const uint32_t q22 = 1 << 22;
+                uint32_t *pw = iq22Payload;
+                uint32_t *pp = (uint32_t*)(paramUpdPayload + offset);
+                *pw++ = (((double)*pp++) / 100 * q22);
```
rotorflight, line 1791: Why is double used here? First, STM32F7/F4 does not support double, so this would be enumlated.
Second. I am not sure why use double/float at all, as this is all integers?

src/main/sensors/esc_sensor.c
```diff
@@ -1518,6 +1440,658 @@ static void apdSensorProcess(timeUs_t currentTimeUs)
...
+        const uint16_t *pal = tribParamAddrLen + i;
+        const uint8_t len = *pal & TRIB_PAL_LEN_MASK;
+        if ((*pal >> 8) == addr) {
+            // cache params by addr
+            if (addr == TRIB_PARAM_IQ22_ADDR) {
+                // convert IQ22 -> scaled uint32
+                const uint32_t q22 = 1 << 22;
+                uint32_t *pr = (uint32_t*)(buffer + TRIB_HEADER_LENGTH);
+                uint32_t *pp = (uint32_t*)(paramPayload + offset);
+                *pp++ = round(((double)*pr++) / q22 * 100);
```
rotorflight, line 1832: Same question about the use of double.

rotorflight (changes requested):

src/main/sensors/esc_sensor.c
bob01: Possibly - docs say ...
 *     24:      Throttle % (-100..100)
 ie apply offset for 0-200, then scale *5 to be equivalent to .pwm 0..100 (*10)
     escSensorData[0].throttle = (throttle + 100) * 5;
     escSensorData[0].pwm = pwm * 10;

src/main/sensors/esc_sensor.c
bob01: Agreed - done.

src/main/sensors/esc_sensor.c
bob01: done.

src/main/sensors/esc_sensor.c
bob01: Evolutionary leftovers - collecting as 32 done.

src/main/sensors/esc_sensor.c
bob01: done.

src/main/sensors/esc_sensor.c
bob01: done.

src/main/sensors/esc_sensor.c
bob01: IQ22 float conversion required better precision than w/ integer only math.
test - expected is 150
    const uint32_t q22 = 1 << 22;
    uint32_t iqv = 0x00001893; // <= from ESC
    uint32_t d1= round(((double)iqv) * 100000 / q22);
    uint32_t d2 = ((uint32_t)iqv) * 100000 / q22;
    printf("\niq22> %d, %d", d1, d2);
output
    iq22> 150, 149

src/main/sensors/esc_sensor.c
bob01: As above. 
These are the gov P and I parameters - one option would be not to present them but some like to use the ESC's internal governor and expect to be able to tweak these. Odd choice by Scorpion.

src/main/sensors/esc_sensor.c
rotorflight: You could do it correctly with integers, but it would look ugly.
Please use float instead. Double is trouble.

rotorflight (changes requested):
Please use float instead of double

src/main/sensors/esc_sensor.c
bob01: Verifying use of float w/ unit tests prior to change...

bob01: float passes tests - float replaces double

src/main/sensors/esc_sensor.c
bob01: float replaces double

src/main/sensors/esc_sensor.c
```diff
@@ -1518,6 +1439,658 @@ static void apdSensorProcess(timeUs_t currentTimeUs)
...
+        const uint8_t len = *pal & TRIB_PAL_LEN_MASK;
+        if ((*pal >> 8) == addr) {
+            // cache params by addr
+            if (addr == TRIB_PARAM_IQ22_ADDR) {
+                // convert IQ22 -> scaled uint32
+                const uint32_t q22 = 1 << 22;
+                uint32_t *pr = (uint32_t*)(buffer + TRIB_HEADER_LENGTH);
+                uint32_t *pp = (uint32_t*)(paramPayload + offset);
+                *pp++ = round(((float)*pr++) / q22 * 100);
+                *pp = round(((float)*pr) / q22 * 100000);
```
rotorflight, line 1832: round() takes a double argument, causing a promotion float -> double.
You need to use roundf()

rotorflight (changes requested):
Nearly there. You missed the round() function. Use roundf() instead.

rotorflight:
Most floating point functions in C are taking double. Need to use the ...f() variant for floats.
Also, all constants need to be suffixed with f, like 1.234f. Otherwise they are double.
It's easy to miss these... BF code use to be full of unnecessary conversions and slow code.

src/main/sensors/esc_sensor.c
bob01: Ah yes, the joys of implicit coercion round(...) replaced with roundf(...), tests and device tests ok

bob01:
Good catch - thanks

rotorflight (approved):

bob01:
Yay - thanks.

PR #112, opened 2024-05-07
CMS bugfix and cleanup.

AndersHoglund:
Bugfix profile handling. Was a mixed use of global config items, local statics and dynamics. Profiles did get out of sync if external changes was made, from RFC or via adjustments. Now only picks up current profiles on main profile menu entry, and sticks to that until exit.

Menu string changes for a bit more clarity:
CYCL GAIN - CY XC GAIN etc, to indicate "Cross Coupling" (XC). 
Use TAIL as in RFC, not YAW, in menu heading.

### DIFF
--- a/src/main/cms/cms_menu_imu.c
+++ b/src/main/cms/cms_menu_imu.c
@@ -151,6 +151,8 @@ static const void *cmsx_profileIndexOnChange(displayPort_t *displayPort, const v
 
     pidProfileIndex = tmpPidProfileIndex - 1;
     changePidProfile(pidProfileIndex);
+    pidProfile = pidProfilesMutable(pidProfileIndex);
+    setProfileIndexString(pidProfileIndexString, pidProfileIndex, pidProfile->profileName);
 
     return NULL;
 }
@@ -162,6 +164,8 @@ static const void *cmsx_rateProfileIndexOnChange(displayPort_t *displayPort, con
 
     rateProfileIndex = tmpRateProfileIndex - 1;
     changeControlRateProfile(rateProfileIndex);
+    memcpy(&rateProfile, controlRateProfiles(rateProfileIndex), sizeof(controlRateConfig_t));
+    setProfileIndexString(rateProfileIndexString, rateProfileIndex, controlRateProfilesMutable(rateProfileIndex)->profileName);
 
     return NULL;
 }
--- a/src/main/cms/cms_menu_imu.c
+++ b/src/main/cms/cms_menu_imu.c
@@ -272,20 +272,20 @@ static const OSD_Entry cmsx_menuRateProfileEntries[] =
 {
     { "-- RATE --", OME_Label, NULL, rateProfileIndexString },
 
-    { "RC R RATE",   OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rcRates[FD_ROLL],    1, CONTROL_RATE_CONFIG_RC_RATES_MAX, 1, 10 } },
-    { "RC P RATE",   OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rcRates[FD_PITCH],    1, CONTROL_RATE_CONFIG_RC_RATES_MAX, 1, 10 } },
-    { "RC Y RATE",   OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rcRates[FD_YAW], 1, CONTROL_RATE_CONFIG_RC_RATES_MAX, 1, 10 } },
+    { "RC R RATE",   OME_UINT8,  NULL, &(OSD_UINT8_t) { &rateProfile.rcRates[FD_ROLL],  1, CONTROL_RATE_CONFIG_RC_RATES_MAX, 1 }},
+    { "RC P RATE",   OME_UINT8,  NULL, &(OSD_UINT8_t) { &rateProfile.rcRates[FD_PITCH], 1, CONTROL_RATE_CONFIG_RC_RATES_MAX, 1}},
+    { "RC Y RATE",   OME_UINT8,  NULL, &(OSD_UINT8_t) { &rateProfile.rcRates[FD_YAW],   1, CONTROL_RATE_CONFIG_RC_RATES_MAX, 1 }},
 
-    { "ROLL SUPER",  OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rates[FD_ROLL],   0, CONTROL_RATE_CONFIG_SUPER_RATE_MAX, 1, 10 } },
-    { "PITCH SUPER", OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rates[FD_PITCH],   0, CONTROL_RATE_CONFIG_SUPER_RATE_MAX, 1, 10 } },
-    { "YAW SUPER",   OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rates[FD_YAW],   0, CONTROL_RATE_CONFIG_SUPER_RATE_MAX, 1, 10 } },
+    { "ROLL SUPER",  OME_UINT8,  NULL, &(OSD_UINT8_t) { &rateProfile.rates[FD_ROLL],    0, CONTROL_RATE_CONFIG_SUPER_RATE_MAX, 1 }},
+    { "PITCH SUPER", OME_UINT8,  NULL, &(OSD_UINT8_t) { &rateProfile.rates[FD_PITCH],   0, CONTROL_RATE_CONFIG_SUPER_RATE_MAX, 1 }},
+    { "YAW SUPER",   OME_UINT8,  NULL, &(OSD_UINT8_t) { &rateProfile.rates[FD_YAW],     0, CONTROL_RATE_CONFIG_SUPER_RATE_MAX, 1 }},
 
-    { "RC R EXPO",   OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rcExpo[FD_ROLL],    0, 100, 1, 10 } },
-    { "RC P EXPO",   OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rcExpo[FD_PITCH],    0, 100, 1, 10 } },
-    { "RC Y EXPO",   OME_FLOAT,  NULL, &(OSD_FLOAT_t) { &rateProfile.rcExpo[FD_YAW], 0, 100, 1, 10 } },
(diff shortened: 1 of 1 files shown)


rotorflight (changes requested):
If I understood this correctly, there is a possible issue with the PID profile init.
If the PID profile has changed, `pidInitProfile(pidProfile);` will load incorrect values from the old profile.

I think you should check first if the system is still using the same profile that CMS is editing, and only then call pidInitProfile.

rotorflight:
And to make it even more bulletproof, make do
```
if (profile-not-changed)
    pidInitProfile(currentPidProfile);
```

i.e. use currentPidProfile pointer instead.

AndersHoglund:
Well, I have tested this case. i.e. modifying items in profile 1 and selecting profile 2 via adjustments. and CMS saved to profile 1. 
Using the externally modifiable "currentPidProfile" and "getCurrentPidprofile" was part of the problem when used together with the CMS locally cached pidProfile and pidProfileIndex. Mixed usage of globals and locals that are out of sync. Now we ONLY work with locally cached profiles, initialized ONCE at entry. At exit all is written back based on the local pidProfile, NOT the global currentPidProfile that might have been changed externally. Your suggestion would bring back the original problem.

Weather or not profiles and other items are to be saved are selected at exit, "exit" or "save&exit". Not exactly sure how this is conceived to work since there are only one "onExit" method. Multiple layers of config storage i think. So, as CMS is designed now on menu exit, current runtime profiles will be changed regardless of "exit" or "exit&save". But after a power cycle, old values will be back unless "exit&save" is used. Not 100% sure on this, but that is how I interpret my tests. More tests needed I think. This could be regarded as a bug too, but a case of it's own. Not directly connected to this PR.

AndersHoglund:
Poor testing. One case missed where profile index change by menu selection did not update properly. Fix in commit bd479fd.

rotorflight (changes requested):
You must not call `pidInitProfile()` with a pointer that isn't the current profile.
(it should not have that argument at all... it's a BF relic.) It must be called always with the currently active profile.
If I am reading this correctly, you are caching the current profile pointer in `pidProfile` and in the end calling `pidInitProfile(pidProfile)`. 
This will break many things if the `pidProfile != currentPidProfile`

AndersHoglund:
The intention is to update and re-init the cached profile we are working with in CMS, NOT the current profile which may have changed externally. The old code did overwrite the wrong profile in this way, I tried to fix that by NOT re-init the current profile. But I'll re-review the code and see if I might have misunderstood the old BF thinking.

rotorflight:
`pidInitProfile()` is for activating the new profile when the profile is changed.
Or if the parameters are changed in the currently active profile.
(The BF interface for changing profiles is poorly designed, and is confusing)

You can save the edited values into the cached profile pointer - that's totally fine.
But you can call `pidInitProfile()` only if the `currentPidProfile == pidProfile`

AndersHoglund:
Well, that is the though; CMS switching active profile to the one being edited and updating the actual pid items. If this is not done, the edited items would not be in effect until saved, a reboot is done. This is connected to the issue of how menus are exited, saved  and system rebooted. I am looking into the possible need to add the REBOOT_REQUIRED flag.
Then the only option is to NOT call pidInitProfile at all from CMS and add this REBOOT_REQUIRED flag. Will look into that and re-test. Not sure this will fix the different handling if backing out from a sub-menu or doing a direct exit. Might need a new menu state flag, a dirty bit. And maybe a state machine around that.....phew. 
Not right now though, flying weather outside.....

rotorflight:
I am afraid you can't switch the active profile in CMS. If the profile has changed, it is because the adjustment functions has changed it, which means it would cause a conflict.

rotorflight:
changing PID profile parameters in RF does not require reboot (unlike BF).
You can freely change them, and they will be used whenever the user changes to that profile later.

rotorflight:
Also, you don't need to save for the new values to take effect. Saving them just means they are still there after reboot.

AndersHoglund:
OK, then it seems correct for CMS to call pidInitProfile as is now, taking effect directly. Either this or reboot I think.
I insist it was wrong to call pidInitProfile with the currentPidProfile arg, caused over-writing the wrong profile if profile index was changed externally. Using the locally cached pidProfile is more correct. Either this or not at all.

AndersHoglund:
Reviewing the code I see what you mean and you are absolutely correct. Calling pidInitProfile is not the correct way, instead changePidProfile should be used. This function makes sure pid controllers currentPidIndex and the actual pid item values are in sync. Commit 7ad20af
.Also, this should probably not be done if no changes are made. Will need figure out how to fix this, not sure how yet. If required (?).

src/main/cms/cms_menu_imu.c
```diff
@@ -190,7 +205,7 @@ static const void *cmsx_PidWriteback(displayPort_t *pDisp, const OSD_Entry *self
         pidProfile->pid[i].O = tempPidO[i];
         pidProfile->pid[i].B = tempPidB[i];
     }
-    pidInitProfile(currentPidProfile);
+    changePidProfile(pidProfileIndex);
```
rotorflight, line 208: You MUST NOT try to change the PID profile from CMS.
It will go out-of-sync with the Tx control.

src/main/cms/cms_menu_imu.c
```diff
@@ -331,21 +332,19 @@ static const void *cmsx_profileYawOnExit(displayPort_t *pDisp, const OSD_Entry *
...
-    pidInitProfile(currentPidProfile);
-
     pidProfile->yaw_cw_stop_gain =       cmsx_yawStopCW;
     pidProfile->yaw_ccw_stop_gain =      cmsx_yawStopCCW;
     pidProfile->yaw_cyclic_ff_gain =     cmsx_yawCyclicFF;
     pidProfile->yaw_collective_ff_gain = cmsx_yawCollectiveFF;
     pidProfile->governor.tta_gain =      cmsx_yawTTA;
     pidProfile->yaw_precomp_cutoff =     cmsx_yawPrecompCutoff;
 
+    changePidProfile(pidProfileIndex);
```
rotorflight, line 342: Ditto. Can't do this.

src/main/cms/cms_menu_imu.c
```diff
@@ -410,14 +402,15 @@ static const void *cmsx_profileCtrlOnExit(displayPort_t *pDisp, const OSD_Entry
       pidProfile->iterm_relax_cutoff[i] = cmsx_pidItermRelaxCutoff[i];
     }
 
+    changePidProfile(pidProfileIndex);
```
rotorflight, line 405: Ditto.

src/main/cms/cms_menu_imu.c
```diff
@@ -468,19 +457,13 @@ static const void *cmsx_profileCtrlBwOnExit(displayPort_t *pDisp, const OSD_Entr
...
-    pidProfile->cyclic_cross_coupling_ratio = cmsx_pidCyclicCrossCouplingRatio;
-    pidProfile->cyclic_cross_coupling_cutoff= cmsx_pidCyclicCrossCouplingCutOff;
-
     for (uint8_t i = 0; i < 3; i++) {
       pidProfile->gyro_cutoff[i] =  cmsx_pidBandwidth[i];
       pidProfile->dterm_cutoff[i] = cmsx_pidDCutoff[i];
       pidProfile->bterm_cutoff[i] = cmsx_pidBCutoff[i];
     }
 
+    changePidProfile(pidProfileIndex);
```
rotorflight, line 466: Ditto.

src/main/cms/cms_menu_imu.c
```diff
@@ -541,14 +520,12 @@ static const void *cmsx_profileLevelOnExit(displayPort_t *pDisp, const OSD_Entry
...
 
-    pidProfile_t *pidProfile = pidProfilesMutable(pidProfileIndex);
-    pidInitProfile(currentPidProfile);
-
     pidProfile->angle.level_strength =   cmsx_angleStrength;
     pidProfile->angle.level_limit =      cmsx_angleLimit;
     pidProfile->horizon.level_strength = cmsx_horizonStrength;
     pidProfile->horizon.transition =     cmsx_horizonTransition;
 
+    changePidProfile(pidProfileIndex);
```
rotorflight, line 528: ...

src/main/cms/cms_menu_imu.c
```diff
@@ -619,6 +589,7 @@ static const void *cmsx_profileRescueOnExit(displayPort_t *pDisp, const OSD_Entr
     pidProfile->rescue.level_gain =         cmsx_rescueLevelGain;
     pidProfile->rescue.flip_gain =          cmsx_rescueFlipGain;
 
+    changePidProfile(pidProfileIndex);
```
rotorflight, line 592: ...

src/main/cms/cms_menu_imu.c
```diff
@@ -679,16 +646,14 @@ static const void *cmsx_profileGovernorOnExit(displayPort_t *pDisp, const OSD_En
...
-    pidInitProfile(currentPidProfile);
-
     pidProfile->governor.headspeed = cmsx_govHeadspeed;
     pidProfile->governor.gain =      cmsx_govMasterGain;
     pidProfile->governor.p_gain =    cmsx_govP;
     pidProfile->governor.i_gain =    cmsx_govI;
     pidProfile->governor.d_gain =    cmsx_govD;
     pidProfile->governor.f_gain =    cmsx_govF;
 
+    changePidProfile(pidProfileIndex);
```
rotorflight, line 656: ...

rotorflight (changes requested):

rotorflight:
Okay. You can't change PID profile from the CMS. The only way at the moment is the adjustment function, which is associated to a Tx channel controlling it. 

You are free to store the adjusted parameters to the cached PID profile pointer.
Then, you need to do the following:
- If PID profile has changed (by adjfunc), DO NOTHING
- If PID profile is the same the CMS has edited, call `pidInitProfile(currentPidProfile)`

In other words, if you are editing the active PID profile, and it is still active when the CMS has saved the new values, only then you need to call `pidInitProfile`. Otherwise you should not.

AndersHoglund:
Profile change from CMS was visible in RFC. So at least CMS and RFC are in sync. But OK, if adj function does not, then we can not change profile from CMS. But then the CMS profile selection entry will have no other effect than what profile CMS will work on, no actual profile selection. Or should we allow profile selection only there, but not when working on profile items in CMS? I can see a use case for selecting profile via CMS, if user has no other way to change in the field. Lack of Tx channels/switches for the adj function for example. How are LUA dealing with this?

rotorflight (approved):

PR #116, opened 2024-05-17
Added adjustments for accelerometer trim

egonl:
Added two adjustments for trimming accelerometer pitch and roll.

image


src/main/fc/rc_adjustments.h
```diff
@@ -131,6 +135,7 @@ typedef enum {
     ADJUSTMENT_TYPE_PROF  = BIT(1),
     ADJUSTMENT_TYPE_GOV   = BIT(2),
     ADJUSTMENT_TYPE_MIX   = BIT(3),
+    ADJUSTMENT_TYPE_ACC   = BIT(4),
```
rotorflight, line 138: TYPE_NONE is for the parameters that don't need initialisation.
Don't add a new one for the purpose.

src/main/fc/rc_adjustments.c
```diff
@@ -155,6 +157,9 @@ static const adjustmentConfig_t adjustmentConfigs[ADJUSTMENT_FUNCTION_COUNT] =
     ADJ_CONFIG(CROSS_COUPLING_GAIN,     PROF,  0, 250),
     ADJ_CONFIG(CROSS_COUPLING_RATIO,    PROF,  0, 200),
     ADJ_CONFIG(CROSS_COUPLING_CUTOFF,   PROF,  1, 250),
+
+    ADJ_CONFIG(ACC_TRIM_PITCH,          ACC,   -300, 300),
+    ADJ_CONFIG(ACC_TRIM_ROLL,           ACC,   -300, 300),
```
rotorflight, line 162: Use NONE type.

rotorflight (changes requested):

src/main/fc/rc_adjustments.h
egonl: Ok, good to know. It has been fixed.

src/main/fc/rc_adjustments.c
egonl: Done, thanks.

rotorflight (approved):

PR #121, opened 2024-06-16
Corrected CMS copy profiles, was hardcoded for 3 profiles only.

AndersHoglund:
CMS copy profiles was hard coded for 3 profiles only. Now follows config item PROFILE_COUNT, up to max 6 profiles.

### DIFF
--- a/src/main/cms/cms_menu_imu.c
+++ b/src/main/cms/cms_menu_imu.c
@@ -886,15 +886,31 @@ static CMS_Menu cmsx_menuFilterPerProfile = {
 static uint8_t cmsx_dstPidProfile;
 static uint8_t cmsx_dstControlRateProfile;
 
-static const char * const cmsx_ProfileNames[] = {
+#if PID_PROFILE_COUNT >= CONTROL_RATE_PROFILE_COUNT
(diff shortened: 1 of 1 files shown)


src/main/cms/cms_menu_imu.c
```diff
@@ -886,15 +886,31 @@ static CMS_Menu cmsx_menuFilterPerProfile = {
...
+#define PROFILE_COUNT CONTROL_RATE_PROFILE_COUNT
+#endif
+
+static const char * const cmsx_ProfileNames[PROFILE_COUNT+1] = {
     "-",
     "1",
     "2",
     "3"
+#if PROFILE_COUNT >= 4
+    ,"4"
```
rotorflight, line 901: I suggest using "...", (comma in the end for all entries)
as it is more readable. C allows the comma after the last entry, so that's not a problem.

rotorflight: No need to figure out the max number of profiles... you could simply define the names for 6, as thats the maximum

rotorflight (changes requested):

AndersHoglund:
OK, updated according to comment.

rotorflight (approved):

PR #128, opened 2024-07-19
Implemented telem ESCx_MODEL for all supported ESCs

bob01:
- unique ESC signature for all supported ESCs 
Note: not a function of escSensorConfig()->protocol / enum ESC_SENSOR_PROTO which may change over time as devices added/removed

Signatures (first byte of tele data frame signature in most cases)...
#define ESC_SIG_NONE              0x00
#define ESC_SIG_BLHELI32          0xC8
#define ESC_SIG_HW4               0x9B
#define ESC_SIG_KON               0x4B
#define ESC_SIG_OMP               0xD0
#define ESC_SIG_ZTW               0xDD
#define ESC_SIG_APD               0xA0
#define ESC_SIG_PL5               0xFD
#define ESC_SIG_TRIB              0x53
#define ESC_SIG_OPENYGE           0xA5

#define ESC_SIG_RESTART           0xFF          // <== sent if ESC needs to be restarted (some do after changing ESC settings)


rotorflight:
Merged manually

PR #129, opened 2024-07-21
SBUS2 Telemetry output

mmosca:
Work in progress.
Porting my changes from INAV.

RX needs to be plugged in the FPORT plug, not SBUS plug.

* [ ] Define telemetry layout
* [ ] Fill in telemetry sensor info
* [ ] Test
* [ ] Document sensor layout


mmosca:
Looks like syncing master messed things up. I will probably setup a different branch and pr.

PR #130, opened 2024-07-22
Add SBUS2 Telemetry support

mmosca:
This is a port of my INAV code to RotorFlight.

SBUS2 port can be connected to FPORT on rotorflight flight controllers, or on the tx pin of any uart on F7/H7 mcus.

These are the available sensors:

| Slot | Sensort Type | Notes |
| --- | --- | --- |
| 1 | Voltage | FC Voltage sensor. Pack and cell voltages |
| 3 | Current | FC Current sensor. |
| 6 | RPM sensor | Headspeed |
| 7 | Temperature | MCU Core temp |
| 8 | Kontronic ESC | ESC telemetry data |

image


src/main/rx/sbus.c
```diff
@@ -134,6 +140,30 @@ static void sbusDataReceive(uint16_t c, void *data)
...
         } else {
             sbusFrameData->done = true;
+
+#if defined(USE_TELEMETRY) && defined(USE_TELEMETRY_SBUS2)
+            switch(c) {
+                case 0x04: // sbus2, first telemetry page
+                case 0x14: // sbus2, second telemetry page
+                case 0x24: // sbus2, second telemetry page
+                case 0x34: // sbus2, second telemetry page
+                    if (c & 0x4)
```
rotorflight, line 150: This condition seems to be true always. Is that correct?

src/main/telemetry/sbus2.c
```diff
@@ -0,0 +1,171 @@
...
+    // 1 slot
+    send_RPM(6, rpm);
+    // 1 slot - esc temp
+    static int change = 0;
+    change++;
+    int delta = change / 10;
+    delta = delta % 20;
+    send_SBS01T(7, temperature);
+
+    if(isEscSensorActive()) {
```
rotorflight, line 104: You need to check that escData != NULL

src/main/fc/tasks.c
```diff
@@ -409,6 +410,10 @@ task_attribute_t task_attributes[TASK_COUNT] = {
 #ifdef USE_CRSF_V3
     [TASK_SPEED_NEGOTIATION] = DEFINE_TASK("SPEED_NEGOTIATION", NULL, NULL, speedNegotiationProcess, TASK_PERIOD_HZ(100), TASK_PRIORITY_LOW),
 #endif
+
+#ifdef USE_TELEMETRY_SBUS2
+    [TASK_TELEMETRY_SBUS2] = DEFINE_TASK("SBUS2_TELEMETRY", NULL, NULL, taskSendSbus2Telemetry, TASK_PERIOD_US(125), TASK_PRIORITY_LOW),
```
rotorflight, line 415: 8kHz task speed is very fast. Is it necessary to run this fast? It might be a problem for F4, maybe even for F7.

rotorflight:
The code in sbus2_sensors.c looks iffy in a few places. Is that your code, or copied from somewhere else?

rotorflight:
Is this code always sending all sensors once enebled?
Can you choose what sensors are used?

mmosca:
> The code in sbus2_sensors.c looks iffy in a few places. Is that your code, or copied from somewhere else?

That is a copy from https://github.com/BrushlessPower/SBUS2-Telemetry

Not all sensors are used. That is just the code to pack the info in different sensor types.

src/main/fc/tasks.c
mmosca: Priority is low. The sbus2 telemetry has somewhat tight timing requirements.

2ms dead time, follow by 8 telemetry slots of ~650us each, but a bit of dead time is needed between frames.

If we lower too much, we may be too late to send telemetry data, but we risk breaking the timing requirements.

It is probably a good idea to limit it to f7 and h7 anyway, since sbus is an inverted protocol and would need extra hardware on f4.

mmosca:
> Is this code always sending all sensors once enebled? Can you choose what sensors are used?

Only the sensors used in taskSendSbus2Telemetry in sbus2.c are sent.

The sensors used are currently hard coded. The futaba side is a bit limited as some sensors sensors use more than one slot and can only be used in some positions. (Esc, for example, takes a full 8 slot window and can only be on slot 8, 16 or 24).

Different radios also may support a smaller subset of sensors.

src/main/rx/sbus.c
mmosca: Good catch.

src/main/telemetry/sbus2_sensors.c
```diff
@@ -0,0 +1,731 @@
...
+
+void send_s1678_current(uint8_t port, uint16_t current, uint16_t capacity, uint16_t voltage)
+{
+   uint16_t value = 0;
+   uint32_t local = 0;
+   uint8_t bytes[2] = {0, 0};
+
+   // CURRENT
+   local = ((uint32_t)current) * 1 ;
+   value = (uint16_t)local;
```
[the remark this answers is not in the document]
mmosca: I agree. :)

Maybe it was to quiet down a compiler warning.

The * 1 does not make much sense. Maybe it was related to unit conversion at some point. I can't really see how casting to uint32_t first would change the result from just casting it to uint16_t directly,

I can simplify it if you want.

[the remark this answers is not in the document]
rotorflight: If the file is directly from the origin, and there are no bugs... maybe just leave it for now.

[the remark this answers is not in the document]
rotorflight: If it turn out later that it's also buggy, then I would suggest cleaning it up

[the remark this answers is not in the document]
mmosca: That is part of the reason I kept it separate. Easier to cross reference in the future.

There are small changes (formatting) due to copying code around and initialization changes due to compiler warnings, but otherwise, the sensor packing code is the same as the original.

rotorflight:
I am not confortable about the SBUS task running at 8kHz. It's almost certainly interfering with the gyro task.
We had to drop the gyro task to 1k on F411 and G474, because they could not even run that.
I strongly believe we can't run this either at 8k on those platforms. Probably max 1k.

I think there are two optons here:
1) find another way to guarantee the timing requirements. Increasing the task speed is not the right approach.
2) if it has to run at 8k, enable this feature only on H7

mmosca:
I am out of town currently. I will run more tests once I am back with hardware access.

rotorflight (changes requested):

src/main/fc/tasks.c
mmosca: I experimented again with this. 8k is required for all sensors to work.

If I lower it to 4k, the last telemetry slotsbecome unreliable and often don't update.

I did lower the priority to LOWEST and limited it to F7, H7 and G4 mcus since the F4 needs extra circuitry to use a half-duplex inverted uart.

mmosca:
I believe I addreessed all items. Let me know if you need any more changes.

mmosca:
Tasks output on flydragonf722_v2 with latest changes and lowest priority.

Realtime load reported at ~26-8% and cpu load ~21-22% on configurator status bar, with all features enabled as well as working sbus2 telemetry,

```
# tasks
Task list             rate/hz  max/us  avg/us maxload avgload  total/ms   late    run reqd/us
00 - (         SYSTEM)     10       5       1    0.0%    0.0%         1      0      7       4
01 - (         SYSTEM)    993       4       1    0.3%    0.1%       124      1    629       1
02 - (           GYRO)   3231       5       2    1.6%    0.7%       820      0   2044       0
03 - (         FILTER)   3231       5       1    1.6%    0.6%       663      0   2044       0
04 - (            PID)   3231      58      47   18.7%   15.4%     17429      0   2044       0
05 - (            ACC)    798       8       3    0.6%    0.2%       260      0    504       5
06 - (       ATTITUDE)    497      13       7    0.6%    0.3%       444      0    314      10
07 - (             RX)     66      18      12    0.1%    0.0%       290      0    126      17
08 - (         SERIAL)     98    1070       2   10.4%    0.0%        75      0     61     219
09 - (       DISPATCH)    993       7       0    0.6%    0.0%        83      0    629       5
10 - (BATTERY_VOLTAGE)    200       3       2    0.0%    0.0%        57      0    126       3
11 - (BATTERY_CURRENT)    200       1       0    0.0%    0.0%        19      0    126       1
12 - ( BATTERY_ALERTS)      5       2       2    0.0%    0.0%         0      0      3       3
13 - (         BEEPER)     98       2       1    0.0%    0.0%        14      0     61       1
16 - (           BARO)     43      29       6    0.1%    0.0%       103      0     72       8
17 - (      TELEMETRY)    249       3       0    0.0%    0.0%       133      0    156       2
18 - (       LEDSTRIP)     98      17       5    0.1%    0.0%        46      0     61      33
21 - (    ADCINTERNAL)      1       3       3    0.0%    0.0%         0      0      1       8
23 - (SBUS2_TELEMETRY)   6345       6       0    3.8%    0.3%       451      0   3986       3
RX Check Function                   1       0                        12
Total (excluding SERIAL) 
```

rotorflight:
Thank you! Merged manually.

PR #131, opened 2024-07-27
Pid bandwidth blackbox headers

rtlopez:
1.  `rollBW`, `pitchBW`, `yawBW` headers
2. missing `f` and `master gain` in `govPID` headers

### DIFF
--- a/src/main/blackbox/blackbox.c
+++ b/src/main/blackbox/blackbox.c
@@ -1480,9 +1480,20 @@ static bool blackboxWriteSysinfo(void)
                                                                             currentPidProfile->angle.level_limit,
                                                                             currentPidProfile->horizon.level_strength,
                                                                             currentPidProfile->horizon.transition);
-        BLACKBOX_PRINT_HEADER_LINE("govPID", "%d,%d,%d",                    currentPidProfile->governor.p_gain,
+        BLACKBOX_PRINT_HEADER_LINE("govPID", "%d,%d,%d,%d,%d",              currentPidProfile->governor.p_gain,
                                                                             currentPidProfile->governor.i_gain,
-                                                                            currentPidProfile->governor.d_gain);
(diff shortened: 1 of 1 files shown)


src/main/blackbox/blackbox.c
```diff
@@ -1480,9 +1480,20 @@ static bool blackboxWriteSysinfo(void)
                                                                             currentPidProfile->angle.level_limit,
                                                                             currentPidProfile->horizon.level_strength,
                                                                             currentPidProfile->horizon.transition);
-        BLACKBOX_PRINT_HEADER_LINE("govPID", "%d,%d,%d",                    currentPidProfile->governor.p_gain,
+        BLACKBOX_PRINT_HEADER_LINE("govPID", "%d,%d,%d,%d,%d",              currentPidProfile->governor.p_gain,
                                                                             currentPidProfile->governor.i_gain,
-                                                                            currentPidProfile->governor.d_gain);
+                                                                            currentPidProfile->governor.d_gain,
+                                                                            currentPidProfile->governor.f_gain,
+                                                                            currentPidProfile->governor.gain);
```
rotorflight, line 1487: Suggest adding another line for the Gov master gain. It would be easier for human reader.

rtlopez: I don' think that many user's read raw blackbox files (with some exceptions like devs), so the data format is not in the first place here. I did it this way to for simplicity, to use function that populate consecutive fields in UI and display in a row. 
bblog_gov
the only one thing I'm concerned, that master gain can be first here like in configurator.

But if you really want, I can extract master gain to separate field.

rotorflight: Fair enough. Just a suggesting. Let's keep it like that

rotorflight (approved):

PR #134, opened 2024-08-06
Etocii testing

rotorflight:
### This is a draft PR for testing new things. NOT to be merged into master.


rotorflight:
New parameters:
- `yaw_precomp_exp`
- `gov_ff_exponent`
- `gov_ff_filter_type`
- `gov_min_pid_throttle`

rotorflight:
### Setpoint acceleration and smoothing

Setpoint acceleration is now indicating how fast the stick can move. The unit is time in ms for the stick to go from 0 to full deflection. For example, 100ms would mean the stick movement is limited to 0.1s from zero to hero.

A human can possibly slam the stick in 10ms if really trying. Normal flying is somewhere around 20ms.
Thus, values > 20 are needed for slowing down the reaction. Values > 100 are likely to feel really slow.

Smoothing is rounding the final approach to the end. Typically a value similar to the accel can be used as a starting point, and then tweaked to liking.

PR #135, opened 2024-08-10
[SBUS2] Fix units reported in esc sensor and report rotorflight's rpm, instead of escData

mmosca:
I could not match  rotorflight's rpm data with headspeed by setting the right gear ratio on the radio.

So it is probably better to set ratio to 1:1 in the radio and report rotorflight's rpm.

### DIFF
--- a/docs/Telemetry.md
+++ b/docs/Telemetry.md
@@ -345,7 +345,7 @@ It runs at a fixed baud rate of 100000 8e2.
 ```
 
 ### Radio Configuration
-The following sensors should setup on your radio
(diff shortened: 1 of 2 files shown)


robthomson:
Can't imagine why you would want to calc it on the radio!

mmosca:
> Can't imagine why you would want to calc it on the radio!

That is just how it is usually done on futaba radios. They just transmit the raw motor rpm sensor output and you would input the type of sensor (hal effect, brushless, motor, optical) and setup as needed.

Letting this be handled by rotorflight makes more sense, since it already has all that information.

rotorflight (approved):

PR #156, opened 2024-09-06
feat: added MSP_PILOT_CONFIG with user definable parameters 

egonl:
The idea is that there will also be a Model page (in Lua/Configurator) where the user can define up to 3 parameters. Those parameters will then be set in rf2bg.lua.

### DIFF
--- a/src/main/cli/settings.c
+++ b/src/main/cli/settings.c
@@ -493,6 +493,10 @@ const char * const lookupTableTelemMode[] = {
     "NATIVE", "CUSTOM",
 };
 
+const char * const lookupTableParamType[] = {
+    "NONE", "TIMER1", "TIMER2", "TIMER3", "GV1", "GV2", "GV3","GV4","GV5","GV6","GV7","GV8","GV9"
+};
+
 #define LOOKUP_TABLE_ENTRY(name) { name, ARRAYLEN(name) }
 
 const lookupTableEntry_t lookupTables[] = {
@@ -604,6 +608,7 @@ const lookupTableEntry_t lookupTables[] = {
     LOOKUP_TABLE_ENTRY(lookupTableSwashType),
     LOOKUP_TABLE_ENTRY(lookupTableDtermMode),
     LOOKUP_TABLE_ENTRY(lookupTableTelemMode),
+    LOOKUP_TABLE_ENTRY(lookupTableParamType),
 };
 
 #undef LOOKUP_TABLE_ENTRY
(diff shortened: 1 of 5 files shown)


src/main/cli/settings.c
```diff
@@ -1701,6 +1706,12 @@ const clivalue_t valueTable[] = {
...
 #endif
     { "model_id",         VAR_UINT8  | MASTER_VALUE, .config.minmaxUnsigned = { 0, 99 }, PG_PILOT_CONFIG, offsetof(pilotConfig_t, modelId) },
+    { "model_param1_type",  VAR_UINT8  | MASTER_VALUE | MODE_LOOKUP, .config.lookup = { TABLE_PARAM_TYPE }, PG_PILOT_CONFIG, offsetof(pilotConfig_t, modelParam1Type) },
+    { "model_param1_value", VAR_INT16  | MASTER_VALUE, .config.minmaxUnsigned = { -32000, 32000 }, PG_PILOT_CONFIG, offsetof(pilotConfig_t, modelParam1Value) },
+    { "model_param2_type",  VAR_UINT8  | MASTER_VALUE | MODE_LOOKUP, .config.lookup = { TABLE_PARAM_TYPE }, PG_PILOT_CONFIG, offsetof(pilotConfig_t, modelParam2Type) },
+    { "model_param2_value", VAR_INT16  | MASTER_VALUE, .config.minmaxUnsigned = { -32000, 32000 }, PG_PILOT_CONFIG, offsetof(pilotConfig_t, modelParam2Value) },
+    { "model_param3_type",  VAR_UINT8  | MASTER_VALUE | MODE_LOOKUP, .config.lookup = { TABLE_PARAM_TYPE }, PG_PILOT_CONFIG, offsetof(pilotConfig_t, modelParam3Type) },
+    { "model_param3_value", VAR_INT16  | MASTER_VALUE, .config.minmaxUnsigned = { -32000, 32000 }, PG_PILOT_CONFIG, offsetof(pilotConfig_t, modelParam3Value) },
```
rotorflight, line 1714: INT16 is not unsigned...
Also, why limit it to -32000...32000 if the data type is not known

rotorflight (changes requested):
Is this fixing the MSP_NAME bug?

egonl:
> Is this fixing the MSP_NAME bug?

There is no MSP_NAME bug, just a poor implementation. Added new MSP_PILOT_CONFIG messages instead of trying to reuse MSP_NAME. - as discussed.

src/main/cli/settings.c
egonl: Thanks! Fixed that.

egonl:
This PR now also fixes the MSP_NAME bug by reverting commit 52237ee5fb56d7facccce99c67fb0ad6aaf97504.

egonl:
Thanks again, fixed!

PR #170, opened 2024-09-24
Allow for longer FC_VERSION_STRING

KryoStoffer:
Make sure buffer is large enough and changed to FC_FIRMWARE_IDENTIFIER, looks slightly better on small screens:


rotorflight:
Fixed in a simpler way 
https://github.com/rotorflight/rotorflight-firmware/commit/67c77895a0bf07350302b130899439d0abadc94d

PR #174, opened 2024-10-02
Add Graupner ESC telemetry support

rotorflight:


src/main/sensors/esc_sensor.c
```diff
@@ -3245,6 +3432,11 @@ bool INIT_CODE escSensorInit(void)
             callback = flySensorInit(escHalfDuplex);
             baudrate = 115200;
             break;
+        case ESC_SENSOR_PROTO_GRAUPNER:
+            callback = graupnerSensorInit();
+            baudrate = 19200;
+            options |= SERIAL_BIDIR;
```
bob01, line 3438: SERIAL_BIDIR option is a function of config->halfDuplex and is already set if configured
... but this protocol appears to function only in request/response mode making half duplex (or full duplex w/ tx/rx) required.
Choices...
- force half duplex operation ON here (now port state does not match port config - may create issues elsewhere)
- enforce config->halfDuplex on setting config->protocol and fail the init (return false if not config->halfDuplex) as a sanity check? (a bit more effort but feels more robust)

rotorflight: I had the second option first, but then I thought what's the point there are no choices here. Why do we want the user to play roulette with it.

rotorflight: The right thing is to set use the checkConfig routine and change the halfDuplex to true if Graupner is selected.

PR #186, opened 2024-10-22
Draft PR for tail testing

rotorflight:


rotorflight:
A note on the inertia precomp.

It is doing something only if the headspeed changes. If the motor/esc/governor were ideal, and the headspeed would stay constant, the estimated coll->yaw precomp would be correct. But, if the headspeed drops, the estimated coll->yaw precomp is too much, and needs to be lowered. This is what the inertia precomp does, it adjusts the yaw precomp based on the headspeed change speed. The RPM drop speed indicates how much torque is "missing" from keeping the RPM constant, and thus how much the yaw precomp must be reduced.

To tune it, first tune the coll->precomp with a full battery, and avoid doing tests that drops the headspeed. Then tune the inertia precomp so that the moves that actually drops the headspeed won't affect the tail. This is escpecially visible when the battery is nearly flat, and the throttle saturates at 100%

PR #187, opened 2024-10-25
Update gtest library and update unit tests

gongtao0607:
Update gtest+gmock library to the most recent release (v1.15.2).

Fix some coding issue causing tests not building:
  * Move main/common/platform.h to main/platform.h\
    This fixes the issue that the platform.h included by common/*.c can't be swapped to test/unit/platform.h. (current directory has top precedence in the quote include).
  * Replace all include "types.h" to "common/types.h". This is to avoid including system "types.h"
  * Rename `pid_t` to
  *  `pidData_t` to avoid naming conflict with `pid_t` in stdlib (process id)
  * Other minor fixes
 
Fix some of the tests, disable the build failure and test failure ones


rotorflight (changes requested):
This PR is too large as a single PR.
Many of the changes are not directly related, so they should be in separate PRs.
Especially the ones that are actually changing the source code must be in a different PR than the ones fixing the tests.

rotorflight:
There are lots of changes to the actual source code. Are they really necessary?

gongtao0607:
Yes, they are necessary to make the test build again.
This PR contain 3 commits. The last one is to fix the tests (and touches the actual code). Most of the changes are “types.h” -> “common/types.h”. The reason of each (global) change is listed in the PR description.

gongtao0607:
I played with the Makefile rule to workaround the conflicts. Now I have reverted all except one change to the source code (`main/common/platform.h` to `main/platform.h`, which is required). Please take another look.

gongtao0607:
The following combinations cause the build issue:
  1. The main `platform.h` needs to be swapped by `test/unit/platform.h`
  2. Many `common/xxx` files include the `platform.h` by quote include: `#include "platform.h"`
  3. When using "Quote include", the current directory (the `common/`) has highest precedence, and no flag can override that behavior. [Source](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html)

1 and 3 are contradicting and the easiest way to resolve I think is to move `platform.h` out of `common/`. It doesn't have to be under `main/`. It just needs to be in a directory that files in the same directory do not quote-include it.

There are alternative solutions but I think they are not as clean as simply moving the file:
  1. try to combine test specific `platform.h` to the main `platform.h` and don't swap it in the test. (breaking 1)
  2. Replace all `#include "platform.h"` with `#include <platform.h>`. (breaking 2/3)

LMK

rotorflight:
I renamed the unit/platform.h to unit/unittest_platform.h, and included it from common/platform.h.
Job done.

rotorflight:
Merged manually.

PR #192, opened 2024-11-24
Governor D gain is off by 10x

awigen:


rotorflight (approved):

ArchieRC:
I've seen multiple cases where this change is making D oscillate on Gov.    previous gov tunes with values of 10 are now having to be lowered to a value of 1 or 2.    This is reducing the ability to be able to tune the D gain to prevent ripples in the P gain.

PR #195, opened 2024-12-03
FBUS Support

robthomson:
Support for FBUS Protocol.


rotorflight (changes requested):
I think we should call it fbus.c on the top level.
Likewise, unless things

make/source.mk
```diff
@@ -126,6 +126,7 @@ COMMON_SRC = \
             rx/sumh.c \
             rx/xbus.c \
             rx/fport.c \
+            rx/fport2.c \
```
rotorflight, line 129: rx/fbus.c

src/main/build/debug.c
```diff
@@ -60,6 +60,7 @@ const char * const debugModeNames[DEBUG_COUNT] = {
     DEBUG_NAME(MAX7456_SPICLOCK),
     DEBUG_NAME(SBUS),
     DEBUG_NAME(FPORT),
+    DEBUG_NAME(FPORT2),
```
rotorflight, line 63: New debug items _MUST_ go to the end of the list

src/main/build/debug.h
```diff
@@ -71,6 +71,7 @@ typedef enum {
     DEBUG_MAX7456_SPICLOCK,
     DEBUG_SBUS,
     DEBUG_FPORT,
+    DEBUG_FPORT2,
```
rotorflight, line 74: New debug items _MUST_ go to the end of the list

rotorflight (changes requested):
Let's call it fbus, as that's what FrSky is calling it

robthomson:
> Let's call it fbus, as that's what FrSky is calling it

So this is an issue we discussed this end quite a bit.

I can rename it.. But we also have fport 2 as part of the same tree.

And to add to the mix frsky mark their fbus receivers with labels saying fport!

It's  a mess on their naming.


I can for sure change this to be fbus.

I wonder maybe if we do a full split and do fport 2 in its own source files?

src/main/build/debug.h
robthomson: Doe

src/main/build/debug.c
robthomson: Done

src/main/cli/settings.c
```diff
@@ -229,10 +229,12 @@ static const char * const lookupTableSerialRX[] = {
     "CRSF",
     "SRXL",
     "CUSTOM",
-    "FPORT",
+    "FPORT",
```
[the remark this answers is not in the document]
robthomson: Fixed

make/source.mk
robthomson: All done

rotorflight:
Integrated manually

PR #200, opened 2024-12-06
First draft for BMI088 driver

NightHawk32:
BMI088 consists of an ACC and Gyro digital part which are tied to the same SPI bus but require separate CS lines.
This required some hacks, since the current architecture trears the IMU as one device/component. This PR tires to intorduce an andditional CS pin without braking support for the other IMUs.

- added BMI088 driver files
- added cli params for additional CS pin
- set gyro ODR to 2000Hz and ACC ODR to 800Hz

Tested with custom hardware for BMI088 adn RF configurator.

Open tests:
- test in flight
- test of implementation with hardware that offers another sensor (checking that all other IMUs still work fine)


src/main/msp/msp.c
```diff
@@ -1074,21 +1074,23 @@ static bool mspProcessOutCommand(int16_t cmdMSP, sbuf_t *dst)
...
+            if(acc.dev.acc_1G == 2731){
+                scale = 16/3.0;
+            }else if (acc.dev.acc_1G > 512 * 4) {
                 scale = 8;
             } else if (acc.dev.acc_1G > 512 * 2) {
                 scale = 4;
             } else if (acc.dev.acc_1G >= 512) {
                 scale = 2;
-            } else {
+            }else {
```
rotorflight, line 1086: unncessary change

src/main/drivers/accgyro/accgyro_spi_bmi088.h
```diff
@@ -0,0 +1,27 @@
...
+ * will be useful, but WITHOUT ANY WARRANTY; without even the implied
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this software.
+ *
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
rotorflight, line 20: IS this driver going to Betaflight too? If so, then this copyright is ok.
If this is for Rotorflight only, then please use Rotorflight one.

src/main/msp/msp.c
```diff
@@ -1074,21 +1074,23 @@ static bool mspProcessOutCommand(int16_t cmdMSP, sbuf_t *dst)
...
                 scale = 1;
             }
 #endif
 
             for (int i = 0; i < 3; i++) {
 #if defined(USE_ACC)
-                sbufWriteU16(dst, lrintf(acc.accADC[i] / scale));
+                sbufWriteU16(dst, lrintf((int16_t)acc.accADC[i] / scale));
```
rotorflight, line 1093: Not sure what this is for. Can. you explain?

src/main/drivers/accgyro/accgyro_spi_bmi088.h
NightHawk32: Intitially I planned to add the code to betaflight aswell, but I will add the RF copyright for now.

src/main/msp/msp.c
NightHawk32: this corssesponds to the change above, for getting the ACC scaling right in the configurator. In case that is not neededl, both changes are unnecessary.

rotorflight: I am not sure I am following. acc.accADC is a float. The scale variable used to be uint8_t, but you changed it to float. Why do you need to convert accADC to integer before dividing it by the scale?

rotorflight: I would understand if the (int16_t) was before the lrintf() function

NightHawk32: You are right, don't know why I introduced that cast. Doesn't make sense like that in the end.
I changed the scaling to float, since the scaling with +-12G is not an integer anymore.

NightHawk32:
> More comments about the MSP

removed the unneressary cast, is anything else left to be done?

rotorflight (approved):

rotorflight:
Looking good now!

PR #207, opened 2024-12-18
Jeti sensors - RF2.1.X

robthomson:
Added in additional sensors to work with jeti.

Currently added:

headspeed
tailspeed
arming flags
pid profile
rate profile
governor mode
throttle control
adjfunction
adjvalue
esc temp
From tests it seems that this is now the max number of sensors we can add. Next steps would be a total refactor of telem to make sensors selectable. I dont have time to do that - so this will have to suffice.


rotorflight:
Too late for getting included into 2.1.0

robthomson:
No issue.. Will drop this pr

robthomson:
Closing as will not make 2.1.X

PR #208, opened 2024-12-22
Introduce collective scaling

gongtao0607:
Introduce 2 new params `collective_scale_pos` and `collective_scale_neg`.
These 2 params set cyclic -> collective mixing to compensate linkage tilting:
```
SC *= (1 + beta* (SR*SR + SP*SP))
```

To compensate the fish eye ball dropping when cyclic is full, set
positive value to either side (_pos or _neg). To compensate the raising,
set negative value.

`collective_scale_pos` has a default 0 and `collective_scale_neg` has 10 as etocii suggested

MSP MSP_MIXER_CONFIG protocol is updated and corresponding changes needs to be done in the lua and configurator.

### DIFF
--- a/src/main/flight/mixer.c
+++ b/src/main/flight/mixer.c
@@ -346,6 +346,19 @@ static float mixerCollectiveCorrection(float SC)
     return SC;
 }
 
+static float mixerCollectiveScale(float SC, float SR, float SP)
+{
+    float beta;
+    if (SC > 0) {
+        beta = mixerConfig()->collective_scale_pos / 100.0f;
+    } else {
+        beta = -mixerConfig()->collective_scale_neg / 100.0f;
+    }
+    float scale = 1 + beta * (SR * SR + SP * SP);
+    scale = constrainf(scale, 0.0f, 2.0f);
+    return SC * scale;
+}
+
 static void mixerUpdateMotorizedTail(void)
 {
     // Motorized tail control
@@ -415,6 +428,7 @@ static void mixerUpdateSwash(void)
         float TC = mixer.tailCenterTrim;
 
         SC = mixerCollectiveCorrection(SC);
+        SC = mixerCollectiveScale(SC, SR, SP);
 
         SR += mixer.swashTrim[0];
         SP += mixer.swashTrim[1];
--- a/src/main/flight/mixer.c
+++ b/src/main/flight/mixer.c
@@ -350,9 +350,9 @@ static float mixerCollectiveScale(float SC, float SR, float SP)
 {
     float beta;
     if (SC > 0) {
-        beta = mixerConfig()->collective_scale_pos / 100.0f;
+        beta = mixerConfig()->collective_geo_correction_pos / 100.0f;
     } else {
-        beta = -mixerConfig()->collective_scale_neg / 100.0f;
+        beta = -mixerConfig()->collective_geo_correction_neg / 100.0f;
     }
     float scale = 1 + beta * (SR * SR + SP * SP);
     scale = constrainf(scale, 0.0f, 2.0f);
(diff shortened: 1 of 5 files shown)


robthomson:
Do you have a link to the discord thread on this so I can understand what its about?

gongtao0607:
@etocii

etocii:
@robthomson it’s used to solve this problem. Swash drops at low collective in large cyclic conditions. Rolling maneuvers feels more natural after negative scaling is changed to 10 (tested by world class pilots)

https://discord.com/channels/859480656135192579/1284457233178034237/1307441293450018939

src/main/msp/msp.c
```diff
@@ -3042,6 +3044,10 @@ static mspResult_e mspProcessInCommand(mspDescriptor_t srcDesc, int16_t cmdMSP,
         mixerConfigMutable()->swash_tta_precomp = sbufReadU8(src);
         mixerConfigMutable()->swash_geo_correction = sbufReadU8(src);
         mixerInitConfig();
+        if (sbufBytesRemaining(src) >= 2) {
```
rotorflight, line 3047: This should be before mixerInitConfig()

src/main/pg/mixer.c
```diff
@@ -38,6 +38,8 @@ PG_RESET_TEMPLATE(mixerConfig_t, mixerConfig,
     .swash_trim = { 0, 0, 0 },
     .swash_tta_precomp = 0,
     .swash_geo_correction = 0,
+    .collective_scale_pos = 0,
+    .collective_scale_neg = 10,
```
rotorflight, line 42: I am not sure this is consistent between all helicopter models out there. Or is it?

src/main/flight/mixer.c
```diff
@@ -306,6 +306,19 @@ static float mixerCollectiveCorrection(float SC)
...
+static float mixerCollectiveScale(float SC, float SR, float SP)
+{
+    float beta;
+    if (SC > 0) {
+        beta = mixerConfig()->collective_scale_pos / 100.0f;
+    } else {
+        beta = mixerConfig()->collective_scale_neg / 100.0f;
+    }
+    float scale = 1 - beta * (SR * SR + SP * SP);
+    scale = constrainf(scale, 0, 1.0f);
```
rotorflight, line 318: Slightly inconsistent. constrainf takes three float arguments. You are providing an int and a float.
Maybe just use two ints here. Or two floats. Whichever you prefer.

src/main/pg/mixer.c
gongtao0607: @etocii tested and had other folks tested on many different models and it seems consistent (although I don't know why).

src/main/msp/msp.c
gongtao0607: Done.

src/main/flight/mixer.c
gongtao0607: Done

src/main/pg/mixer.c
etocii: I have my Specter, ilGoblin, Nimbus, FW470L all showing very consistent result with the default value of 10.

etocii:
@gongtao0607 After some simulations and more observations, maybe for positive case we could increase collective? so for positive, we can use negative beta value or flip the sign.

rotorflight:
We need to understand why the collective should be adjusted with the square of the cyclic deflection.
And why not linearly? Or are both just approximations?

etocii:
> We need to understand why the collective should be adjusted with the square of the cyclic deflection. And why not linearly? Or are both just approximations?

Of course they are approximations. We need someone to spend more time to make proper incremental measurements from a cad model but from practice it's easy to see that theres are more vertical movement on the swash at the extreme cyclic.

https://github.com/user-attachments/assets/495e88c1-4aff-4ccc-8ae6-a783d27df28d


I believe that the idea is at extreme angles on the servo, the error is related to cos(link tilt), which is very well approximated by 1-beta*x^2

rotorflight:
The servo rotational geometry should be fully corrected by the servo geo correction.
Is this not enough in this case?

gongtao0607:
The PR is reworked so it can adjust both directions.

rotorflight (approved):

Issue #126, opened 2024-07-13
BB log / Auto delete, or overwrite, or rolling log so it is able to record always


azaz44:
it seems BB is now made mainly for tuning, so you can record logs to tune filters or PIDs. But recording means some manual work, at least to erase data so it has space.

However, BB would be a fantastic tool to diagnose issues, in case of heli crash or any incident when reason is unknown. Having access to recent flights at any time would be super handy. Having voltages, currents, but also servo outputs and RF actions, would help to find out very many issues. I had an umber of incidents because of servos getting stuck (because of gears). It took me a lot of time to find reasons. But this would be visible in log files, that RF is trying to push the servo all time and errors in PID accumulate.

Actually, the name BlackBox suggest such usage. Blackboxes in real planes record all the time.
As it is now, if you forget to delete your logs, then it will not log and data is lost.

rotorflight:
This would be great.
This is a large change and will need time to develop. Likely slipping into 2025.

gongtao0607:
This will be nice feature to have so you can set and forget. It will be a fair amount code change:
  1. Make the BBlog storage volume (flashfs.c) a circular volume and update all the references (MSC/emfat and the MSP log dump). (There are some tricky edge cases involved and need to update the existing flashfs framework).
  2. When arming, maintain a certain amount of free space (by erasing a few sectors). This will be usable milestone.
  3. Furthermore, we can support erase a sector in the runtime when needed. From my rough calculation, this approach should be generally fine with NAND flash, whose erase is in a few ms level. On NOR flash, it will need what is called **erase suspend** feature (which is not commonly supported) to compensate the slow erase, and we need to implement the controlling logic.

gongtao0607:
I'm working on a proof of concept "arming erase" and "background erase" logic:
https://github.com/gongtao0607/rotorflight-firmware/tree/flashfs_online_erase

On Flydragon, it works perfectly after increasing the buffer from 128 to 256 (byte) (means no buffer drop). On a Matek however, it has ~10ms buffer drop on every erase. Both have the same flash (NAND) and the actual erase time logged are always around 2ms. I'm still trying to figure out the discrepancy.

I do not have a FC that has NOR flash to test, and I don't expect it to work as of today (without erase suspend).

The arming erase should work regardless. It may locks 10ms to 1s before arming but who cares.

gongtao0607:
[...]

PR #209, opened 2024-12-22
Flashfs Loop recording and initial erase

gongtao0607:
This PR implements: [#126](https://github.com/rotorflight/rotorflight-firmware/issues/126#issuecomment-2521562367) bullet 1 and 2: 

* Flashfs is now a circular volume with a head pointer and tail pointer.
* Initial erase setting allows flashfs to automatically erase a few sectors upon logging start.

`set blackbox_initial_erase = 8388608` (size in bytes) to turn it on, default is 0 (turned off).

Both are tested on Flydragon and Matek G474Heli as my daily firmware

### DIFF
--- a/src/main/blackbox/blackbox.c
+++ b/src/main/blackbox/blackbox.c
@@ -335,6 +335,8 @@ static const blackboxSimpleFieldDefinition_t blackboxSlowFields[] = {
 typedef enum BlackboxState {
     BLACKBOX_STATE_DISABLED = 0,
     BLACKBOX_STATE_STOPPED,
+    BLACKBOX_STATE_WAIT_FOR_READY,
+    BLACKBOX_STATE_INITIAL_ERASE,
     BLACKBOX_STATE_PREPARE_LOG_FILE,
     BLACKBOX_STATE_SEND_HEADER,
     BLACKBOX_STATE_SEND_MAIN_FIELD_HEADER,
@@ -345,6 +347,7 @@ typedef enum BlackboxState {
     BLACKBOX_STATE_CACHE_FLUSH,
     BLACKBOX_STATE_PAUSED,
     BLACKBOX_STATE_RUNNING,
+    BLACKBOX_STATE_FULL,
     BLACKBOX_STATE_SHUTTING_DOWN,
     BLACKBOX_STATE_START_ERASE,
     BLACKBOX_STATE_ERASING,
@@ -1175,7 +1178,7 @@ static void blackboxStart(void)
     blackboxLastRescueState = getRescueState();
     blackboxLastAirborneState = isAirborne();
 
-    blackboxSetState(BLACKBOX_STATE_PREPARE_LOG_FILE);
+    blackboxSetState(BLACKBOX_STATE_WAIT_FOR_READY);
 }
 
 void blackboxCheckEnabler(void)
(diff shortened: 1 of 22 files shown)


rotorflight:
This _really_ needs to get tested with other FCs that have a NOR flash.

gongtao0607:
I do not have any other FCs. Could you help test?

robthomson:
Happy to test.

rotorflight:
This is an absolutely fantastic feature. Let's see if it works with BF boards. Most of them have a NOR flash.
Lots of people are using BF boards, so we can't really kill them off.

gongtao0607:
> Also... the NOR chips are usually much smaller, so another question is if they should be fully erased at arming.
> Typical NOR chips are 64Mbit and 128Mbit, so only 8Mb and 16Mb. Just about enough for one flight.

If NOR flashes erase considerably faster in multi-block erase (or full chip erase) than serialized block erase -- I'm not sure if it's true --, it worth improve the erasing logic. Otherwise, the current logic provides enough flexibility for either full chip, 1/2 chip or whatever threshold erase.

robthomson:
I have one BF board I can test on.   

https://www.jhemcu.com/e_productshow/?73-GF30F722-ICM-73.html

How do I know if it has a NOR chip?

gongtao0607:
> I have one BF board I can test on.
> 
> https://www.jhemcu.com/e_productshow/?73-GF30F722-ICM-73.html
> 
> How do I know if it has a NOR chip?

https://github.com/rotorflight/rotorflight-firmware/blob/master/src/main/target/JHEF7DUAL/target.h#L125
Looks like a 16M M25P16

gongtao0607:
I'm getting very confused:

 * The image of your board clearly shows a Winbond flash chip. In the spec it says 16 Mi**B** flash.
 * The provided config for this board is M25P16
   * M25P16 is a 16Mi**b** (=2Mi**B**) flash chip. Unless it's referring to a general type??
 * In the M25P16 driver it has a list of chips that it can detect. Filter with (Winbond, 16Mi**B** chip), I get W25Q128 [code](https://github.com/rotorflight/rotorflight-firmware/blob/master/src/main/drivers/flash_m25p16.c#L100-L102)
 * In the same folder, a separate w25q128 driver detects the same chip [code](https://github.com/rotorflight/rotorflight-firmware/blob/master/src/main/drivers/flash_w25q128fv.c#L42)

It seems very unorganized.

rotorflight:
I doubt any BF board is using 16 Mbit chips. That's too small even for BF.
Usually they are 64Mbit or 128Mbit

rotorflight:
> It seems very unorganized.

Welcome to the world of BF

gongtao0607:
Worst case NOR users get these advanced features disabled by default but they have nothing to lose. 

They have to figure out a acceptable arm delay v.s. log file size. Or the amount of sensor data to record (data rate) v.s. sector erase speed (background erase).

rotorflight:
I don't think we can leave this to end users to figure out.
It either works with NOR flashes, or the feature should be disabled with them.

gongtao0607:
Tested on flywing f405: it will take 15s to complete a 4MiB initial erase. All other functionalities are normal and uninterrupted. The recording won't start until the erase is complete.

rotorflight (approved):

rotorflight:
Well, that is a rather long time, but the user can choose not to do that.

rotorflight:
I am considering if the `blackbox_initial_erase = 8388608` should be in kilobytes instead of bytes.
People don't like scary large numbers. Especially if they dont understand why it is 8388608 and not 8388609 etc...

gongtao0607:
> I am considering if the `blackbox_initial_erase = 8388608` should be in kilobytes instead of bytes. People don't like scary large numbers. Especially if they dont understand why it is 8388608 and not 8388609 etc... And if it is in kilobytes, maybe the data type can be U16 instead of U32.

[28e92d0](https://github.com/rotorflight/rotorflight-firmware/pull/209/commits/28e92d026e6a709d3470777e13517f546fb27939) (Don't think it saves any byte though.)

rotorflight (approved):

PR #211, opened 2024-12-26
Flashfs rolling erase

gongtao0607:
This PR is based off PR #209. It implements rolling erase.

Background: https://github.com/rotorflight/rotorflight-firmware/issues/126#issuecomment-2521562367 bullet 3

`set blackbox_rolling_erase = ON` to turn it on (default is off).

The feature is tested on Matek G474Heli and Flydragon, both have w25n01g flash chip whose sector erase is ~2ms.

If the sector erase is slower, there will be buffer drops and log gaps, which is generally fine (tested). However, the exact behavior on other flash chips is untested and there may be room for improvements (e.g., erase suspend, larger buffer).

### DIFF
--- a/src/main/blackbox/blackbox.c
+++ b/src/main/blackbox/blackbox.c
@@ -335,6 +335,8 @@ static const blackboxSimpleFieldDefinition_t blackboxSlowFields[] = {
 typedef enum BlackboxState {
     BLACKBOX_STATE_DISABLED = 0,
     BLACKBOX_STATE_STOPPED,
+    BLACKBOX_STATE_WAIT_FOR_READY,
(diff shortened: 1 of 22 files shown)


gongtao0607:
This is rolling erase performed on a flywing F405 with highest blackbox write bandwidth (all sensors enabled and 2000Hz record speed), ~ 86KiB / S.

Screenshot 2025-01-13 233558

The record gap is unfortunately quite large: ~250ms gap per 128k write. But please note this is the highest record setting.

rotorflight (approved):

rotorflight:
Conflict resolved and merged.

Issue #213, opened 2024-12-28
Graceful period between disarm and BB stop recording


gongtao0607:
I always disarm my Heli right before crashing. It will then lose a second of important BB data at crashing. I think we should add a few seconds of graceful period between disarm and stopping the recording. 2 seconds should be enough for most cases.

I know I can use "Switch" logging mode and customize it in the radio to keep it record, or use throttle hold instead of disarm, but I still feel this is a sweet small feature to have.

PR #219, opened 2025-01-06
Blackbox: add 5s graceful period after stopping

gongtao0607:
This is to implement #213 

If logging is turned off from RUNNING state, it will enter a 5s graceful period state and continue log normally.

If logging is turned back on in the graceful period, it will go back to RUNNING state (and form a single BBL).

If logging is turned off from PAUSED state, this PR doesn't alter the behavior.

Test:

Log stopped by disarming:
Screenshot 2025-01-05 224917

Log stopped and restarted within graceful period:
Screenshot 2025-01-05 225004


rotorflight:
Make the delay a parameter

gongtao0607:
> Make the delay a parameter

[b3879ef](https://github.com/rotorflight/rotorflight-firmware/pull/219/commits/b3879efd1149c4ab3cc4da7477ef65c72e5c55ea)

> I think the delay should be on the disarm condition, not on the overall logging end. This would delay the logging stop if triggered by disarm, but not when explicitly turned off by the logging switch.

This will add more complicated logic to `blackboxIsLoggingEnabled` or the state machine. With a parameter delay, I feel this is not necessary comparing to keeping the logic clean. What do you think?

rotorflight:
People requested to add the logic to enable/disable logging with a switch only (without any delays / extra conditions)
Adding the delay would break this functionality. 

Please add the delay to the disarm condition only. It should not be too hard.

gongtao0607:
> People requested to add the logic to enable/disable logging with a switch only (without any delays / extra conditions) Adding the delay would break this functionality.

[8732d28](https://github.com/rotorflight/rotorflight-firmware/pull/219/commits/8732d28b6b50cbaac96aea434c90c167436b78bc)

gongtao0607:
> [8732d28](https://github.com/rotorflight/rotorflight-firmware/commit/8732d28b6b50cbaac96aea434c90c167436b78bc) is not what I asked.
> 
> Please add a delay to the disarm condition, i.e. delay the disarm event.

Not sure if I follow. The new code adds graceful period only on disarm -- (if it would stop recording).

gongtao0607:
OT: we should cherry pick this feature to real blackboxes https://www.nbcnews.com/news/world/jeju-air-black-box-data-missing-last-4-minutes-crash-south-korea-minis-rcna187239

rotorflight:
I can't accept this PR in this form. It is breaking an existing functionality that is important.

Essentially, the logging must be stopped immediately if the logging switch is turned off.
Regardless whether it is in NORMAL mode or SWITCH mode. (obviously not in ARMED mode).

So it should work like this:

ARMED mode:
- turn logging on when armed
- turn logging off after grace period after disarming

SWITCH mode:
- turn logging on / off immediately on the switch

NORMAL mode:
- turn logging on when both armed and switch on
- turn logging off immediately when switch turned off
- turn logging off after grace period after disarming (while switch is still on)

gongtao0607:
Dr Rudder, this PR does exactly what you have described:

ARMED mode:
  * Turn on when armed, by `blackboxIsLoggingEnabled` (No change)
  * Turn off and enter graceful period: `STATE_RUNNING` -> `STATE_GRACEFUL_PERIOD` (Ln.1208)

SWITCH mode:
  * Turn on when switched on, by `blackboxIsLoggingEnabled` (No change) 
  * Turn off immediately: `STATE_RUNNING` -> `STATE_PAUSED` -> `STATE_SHUTTING_DOWN` (Ln.1200) (Essentially no change)

NORMAL mode:
  * Turn on when  both armed and switched on (`blackboxIsLoggingEnabled`, No change) 
  * Pause immediately when switched off: `STATE_RUNNING` -> `STATE_PAUSED` (Ln: 2149) (No change) 
    * If disarm now, it will stop without graceful period: `STATE_PAUSED` -> `STATE_SHUTTING_DOWN` (No change) 
    * If switched on, it will continue logging: `STATE_PAUSED` -> `STATE_RUNNING` (No change) 
  * Turn logging off after grace period after disarming (while switch is still on): `STATE_RUNNING` -> `STATE_GRACEFUL_PERIOD` (Ln. 1208)

pmattila:
Sounds good. I got lost in some of the details.

pmattila (changes requested):
CLI parameter for gracefulPeriod missing

gongtao0607:
> CLI parameter for gracefulPeriod missing

Missed that... Added in [c7afec2](https://github.com/rotorflight/rotorflight-firmware/pull/219/commits/c7afec2ef5f9ce0d1be0c02b8f9e538eeed588c7)

pmattila (approved):

PR #220, opened 2025-01-06
Improved Integral algorithm: offset flood

gongtao0607:
@etocii can give more details about the algorithm. Basically it converts some of the `axisError` to `axisOffset` according to a curve and limit.


etocii:
This algorithm has been tested by many pro pilots and so far people far prefer this one over the old HSI. The idea is that the old algorithm did not charge HSI fast enough (it basically doesn't charge it while cruising). With good conditions such as low cyclic and stable collective, one can assume that all cyclic axisError are caused by wind thus should be charged into axisOffset. 

There are some problems with the decay pattern from the old firmware however, the center decay for axisOffset (offset_decay_rate_curve & offset_decay_limit_curve) was too low to handle the correctly charged offset. at near zero collective, the heli would stop faster than the decay. Thus, the default of center decay is increased. I do not see much negative effect compared to the old decay values.

In addition, we found that the decay limit for axisError (normal I gain) does not need to be increased at high collective. This does not change the transient flight performance but it keeps orientation better when heli is moving at a high speed.

rotorflight (changes requested):
please rename hsflood to offset_flood throughout.
it's HSI == offset so let's keep the naming here.

src/main/pg/pid.c
```diff
@@ -126,6 +126,9 @@ void resetPidProfile(pidProfile_t *pidProfile)
         .governor.collective_ff_weight = 100,
         .governor.max_throttle = 100,
         .governor.min_throttle = 10,
+        .hsflood_curve = { 0,0,0,20,50,100,180,220,220,220,220,220,220,220,220,220 },
+        .hsflood_relax_level = 40,
+        .hsflood_relax_cutoff = 3,
```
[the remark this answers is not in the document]
gongtao0607: Fixed

src/main/cli/settings.c
```diff
@@ -1136,6 +1136,9 @@ const clivalue_t valueTable[] = {
     { "offset_bleed_rate_curve",    VAR_UINT8 | PROFILE_VALUE | MODE_ARRAY, .config.array.length = LOOKUP_CURVE_POINTS, PG_PID_PROFILE, offsetof(pidProfile_t, offset_bleed_rate_curve) },
     { "offset_bleed_limit_curve",   VAR_UINT8 | PROFILE_VALUE | MODE_ARRAY, .config.array.length = LOOKUP_CURVE_POINTS, PG_PID_PROFILE, offsetof(pidProfile_t, offset_bleed_limit_curve) },
     { "offset_charge_curve",        VAR_UINT8 | PROFILE_VALUE | MODE_ARRAY, .config.array.length = LOOKUP_CURVE_POINTS, PG_PID_PROFILE, offsetof(pidProfile_t, offset_charge_curve) },
+    { "hsflood_curve",              VAR_UINT8 | PROFILE_VALUE | MODE_ARRAY, .config.array.length = LOOKUP_CURVE_POINTS, PG_PID_PROFILE, offsetof(pidProfile_t, hsflood_curve) },
```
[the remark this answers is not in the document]
gongtao0607: Fixed

gongtao0607:
[adf4f2a](https://github.com/rotorflight/rotorflight-firmware/pull/220/commits/adf4f2a1bb6e2a00db039adce12e28e5d3685a17)

src/main/flight/pid.c
```diff
@@ -249,6 +249,11 @@ void INIT_CODE pidInitProfile(const pidProfile_t *pidProfile)
     acroTrainerInit(pidProfile);
 #endif
     rescueInitProfile(pidProfile);
+
```
rotorflight, line 252: Move this before "Initialise sub-profiles"

src/main/flight/pid.c
```diff
@@ -978,6 +983,57 @@ static void pidApplyCyclicMode3(uint8_t axis, const pidProfile_t * pidProfile)
...
+            outputDelta = errorDelta * Ki;
+            offsetDelta = outputDelta / collective / Ko;
+        }
+
+        // 3. Update axisError and axisOffset
+        pid.data[axis].axisError -= errorDelta;
+        pid.data[axis].I -= errorDelta * Ki;
+        pid.data[axis].axisOffset += offsetDelta;
+    }
+
```
rotorflight, line 1036: Move this block into a separate function pidApplyOffsetFlood(), just like pidApplyOffsetBleed()

src/main/flight/pid.c
```diff
@@ -1130,6 +1186,15 @@ static void pidApplyYawMode3(void)
...
 }
 
+void pidCalcOffsetFloodRelaxFactor(const pidProfile_t *profile)
+{
+    const float collective = getCollectiveDeflection();
+    const float collectiveLpf =
+        pt1FilterApply(&pid.offsetFloodRelaxFilter, collective);
+    const float collectiveHpf = collective - collectiveLpf;
+    pid.offsetFloodRelaxFactor =
+        MAX(0, 1.0f - fabsf(collectiveHpf) / profile->offset_flood_relax_level);
```
rotorflight, line 1196: This function should be part of pidApplyOffsetFlood()

rotorflight (changes requested):
Some code formatting things left.
Trying to keep the pid.c tidy.

src/main/flight/pid.c
gongtao0607: Done

src/main/flight/pid.c
gongtao0607: Done

src/main/flight/pid.c
gongtao0607: Done

src/main/flight/pid.c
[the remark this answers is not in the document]
gongtao0607: Fixed both

gongtao0607:
[feb16d6](https://github.com/rotorflight/rotorflight-firmware/pull/220/commits/feb16d68fd8f9511261f3bdfdab0fd891ab626eb)

rotorflight (approved):

PR #221, opened 2025-01-11
Use Push-Pull drive on S.Port/F.Port/FBUS

rotorflight:
S.Port is an inverted signal, with a pull-down resistor. If un-inverted signalling is preferred, it can't be used with on OD output, but requires a PP output. Otherwise there is nothing to pull up the signal level.


robthomson:
Sofar my tests have shown no issues.

BladeScraper-Designs:
No issues so far on tested Nexus + Archer Plus RSMini, TDMX both inverted and uninverted output

PR #224, opened 2025-01-16
Flashfs: erase suspend for w25q flash chips

gongtao0607:
When using rolling erase, all the writes have to wait (buffered) for the erase to complete. If erase is too slow, the buffer overflows and writes get dropped, which will result in gaps in the log:

Screenshot 2025-01-13 233558

This is a log from flywing f405 (w25q128 flash chip) running rolling erase. The erase time is ~120ms and we simply cannot afford that large buffer (we will need to buffer > 6KiB data).

Another way to eliminate the gap without large buffer is to "suspend" the erase when the write comes. This is a function **some** flash chips support. -- Suspend an ongoing erase so you can temporarily perform write/read, or suspend an ongoing write so you can temporarily perform read. (w25q family supports it.)

Key logic:
  * When `FlushAsync` triggers, if a rolling erase is running and buffer is half filled, send `suspend` cmd.
  * On w25q chips, `suspend` cmd takes 20us to process. This is a bit costly to busy waiting, so we add it as a state.
  * If suspend is complete (`SUS` status bit set), flashfs moves to a SUSPENDED state.
  * When the next `FlushAsync` triggers, it should proceed to `Program` (instead of bailing out).
  * Once the flashfs write buffer is below the threshold (half), send `resume` cmd and move back to ROLLING_ERASING state
 
~~Some considerations: `FlushAsync` frequency limits the state changing efficiency. Comparing to a normal write, a write-suspend-write-again takes extra 'EraseAsync' task period (1ms) +'FlushAsync' task period (1ms) to complete (at least). Another 'EraseAsync' task period (1ms) to resume the erase. We lose those time (flash bandwidth) neither erasing or writing. -- basically 3ms wasted for each suspend.~~

This PR is tested on w25n01g (no change) and w25q128 (erase suspend).

This is a capture from flywing f405 (w25q128), rolling erase with erase suspend:
image

https://github.com/rotorflight/rotorflight-firmware/pull/224#issuecomment-2601673305 for a better view of erase/suspend/write switching


gongtao0607:
After increase flashfs task frequency (1kHz -> 2kHz), the performance is largely improved and no gaps in a 1min recording.

gongtao0607:
This feature has been tested on flydragon, and flywing f405 (flight test).

Bench tested on flydragon, matek g474 and flywing f405.

gongtao0607:
A screen capture of erase suspend/resume events logged in the blackbox log (debug build).

image

Explanation: we are doing (switched) erase while writing. The block erase completes within roughly 1/4 of the time we writing the whole block. (Therefore the erase is fast enough, even switched).

rotorflight:
Please add description what improvements this feature is giving. Or which issue it is fixing.

rotorflight:
Have you checked what the PID loop executiong times are like with this?
That's the debug mode PIDLOOP. If the excution time of any step is getting closer to 100us, we risk loosing gyro samples.
It is more important not to loose gyro samples, rather than have issues with the BB.

rotorflight:
It's taking a bit longer to get this merged, as I am testing this against the new driver I have been working on. Stay tuned.

rotorflight (approved):

PR #226, opened 2025-01-20
Add JR DMSS Xbus Mode A protocol into RF

TheNige069:
The additions are to add the JR XBus Mode A protocol to the list of available protocols in Rotorflight. This will allow additional channels, initially set to 16,  to be used for adjustments, as well as higher resolution for servo control


src/main/rx/xbus.c
```diff
@@ -80,17 +87,89 @@
...
 static uint8_t xBusProvider;
 
-
 // Use max values for ram areas
-static volatile uint8_t xBusFrame[XBUS_FRAME_SIZE_A2];  //size 35 for 16 channels in xbus_Mode_B
-static uint16_t xBusChannelData[XBUS_RJ01_CHANNEL_COUNT];
+static volatile uint8_t xBusFrame[XBUS_MODEA_FRAME_SIZE];  // size 69 for 16 channels in xbus_Mode_A
+static uint16_t xBusChannelData[XBUS_MODEA_CHANNEL_COUNT];
+
+// xBus Mode A uses an 8-bit CRC based on Dallas/Maxim
```
rotorflight, line 119: Please don't use large data tables, they flash space is precious.
Instead, use the CRC functions in common/crc.h instead.

src/main/rx/xbus.c
```diff
@@ -112,6 +191,61 @@ static uint8_t xBusRj01CRC8(uint8_t inData, uint8_t seed)
...
+                // Byte 1 - CH Function (unused so ignore)
+                // Byte 2 - CH Data-High
+                // Byte 3 - CH Data-Low
+                // Channel number is 0 based, so we need to subtract 1 from the value
+                const uint8_t nChannelNumber = xBusFrame[offsetBytes + i * 4] - 1;
+                const uint8_t frameAddr = offsetBytes + 2 + i * 4;
+                uint16_t value = ((uint16_t)xBusFrame[frameAddr]) << 8;
+                value += ((uint16_t)xBusFrame[frameAddr + 1]);
+
+                if (value >= 0xFFFF) //65535
```
rotorflight, line 234: uint16 can't be larger than 0xFFFF. Either it should be uint32_t, or this test is meaningless.

src/main/rx/xbus.c
```diff
@@ -197,26 +331,36 @@ static void xBusDataReceive(uint16_t c, void *data)
     now = micros();
```
rotorflight, line 331: Please use the time data types and cmp32() for comparing them.
Otherwise the wraparound will cause issues.

src/main/rx/xbus.c
```diff
@@ -112,6 +191,61 @@ static uint8_t xBusRj01CRC8(uint8_t inData, uint8_t seed)
...
+                // Channel packets are constructed as such:
+                // Byte 0 - Channel number
+                // Byte 1 - CH Function (unused so ignore)
+                // Byte 2 - CH Data-High
+                // Byte 3 - CH Data-Low
+                // Channel number is 0 based, so we need to subtract 1 from the value
+                const uint8_t nChannelNumber = xBusFrame[offsetBytes + i * 4] - 1;
+                const uint8_t frameAddr = offsetBytes + 2 + i * 4;
+                uint16_t value = ((uint16_t)xBusFrame[frameAddr]) << 8;
+                value += ((uint16_t)xBusFrame[frameAddr + 1]);
```
rotorflight, line 232: Should be or | not plus.

rotorflight (changes requested):

src/main/rx/xbus.c
TheNige069: I have updated the code to use time data type (timeUs_t).
I have not changed the code to use cmp32 as doing so causes a compile time error message stating that xBusTimeLast is undeclared, even though it is declared a couple of lines before.
As cmp32 will expand to "a - b" at compile time, the below statement is essentially the same.
    xBusTimeInterval = now - xBusTimeLast;
Also, this code is as written in Cleanflight/Betaflight when XBus Mode B was originally added, back in 2017 I think it was, then later the RJ01 code was added

src/main/rx/xbus.c
TheNige069: Changed to |= from +=

src/main/rx/xbus.c
TheNige069: Lines removed.

src/main/rx/xbus.c
```diff
@@ -80,17 +87,55 @@
...
-static volatile uint8_t xBusFrame[XBUS_FRAME_SIZE_A2];  //size 35 for 16 channels in xbus_Mode_B
-static uint16_t xBusChannelData[XBUS_RJ01_CHANNEL_COUNT];
+static volatile uint8_t xBusFrame[XBUS_MODEA_FRAME_SIZE];  // size 69 for 16 channels in xbus_Mode_A
+static uint16_t xBusChannelData[XBUS_MODEA_CHANNEL_COUNT];
+
+// xBus Mode A uses an 8-bit CRC based on Dallas/Maxim
+// Will use one of the CRC functions in common/crc.c once I can get it working
+// Currently when using a poly of 0x31 and base of 0x00, the wrong values are being
+// returned.
+uint8_t crc8_dallas(const uint8_t* data, const uint8_t size)
```
rotorflight, line 123: Should be static if only used in this file

src/main/rx/xbus.h
```diff
@@ -20,4 +20,5 @@
 
 #pragma once
 
+uint8_t crc8_dallas(const uint8_t* data, const uint8_t size);
```
rotorflight, line 23: Should not be here if only used in xbus.c

rotorflight (changes requested):
A few minor things

src/main/rx/xbus.c
TheNige069: Yes, it should be static as a this stage it is only used in the XBus files, but I'm planning ahead as the telemetry also uses the same CRC calculation and I am currently working on integrating JR telemetry.

src/main/rx/xbus.h
TheNige069: As per my other comment.  Planning ahead.

src/main/rx/xbus.c
```diff
@@ -190,33 +287,45 @@ static void xBusDataReceive(uint16_t c, void *data)
...
     // Check if we shall reset frame position due to time
-    now = micros();
-    xBusTimeInterval = now - xBusTimeLast;
+    const timeUs_t now = microsISR();
+    //xBusTimeInterval = now - xBusTimeLast;
+    xBusTimeInterval = cmp32(now, xBusTimeLast);
     xBusTimeLast = now;
-    if (xBusTimeInterval > XBUS_MAX_FRAME_TIME) {
+
+    if (xBusTimeInterval > (timeDelta_t)xBusMaxFrameTime) {
```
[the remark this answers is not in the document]
TheNige069: Changed declaration to timeDelta_t.

src/main/io/serial.h
```diff
@@ -52,6 +52,7 @@ typedef enum {
     FUNCTION_FRSKY_OSD           = (1 << 16), // 65536
     FUNCTION_DPRINTF             = (1 << 17), // 131072
     FUNCTION_SBUS_OUT            = (1 << 18), // 262144
+    FUNCTION_TELEMETRY_JRDMSS    = (1 << 19), // 524288
```
rotorflight, line 55: Where is this used? Presumably planning ahead again.
I think this should be added when your telemetry stuff is ready.
Not yet here.

rotorflight (changes requested):

src/main/io/serial.h
TheNige069: Agreed. I have removed the offending line and will add it back in under the (to be created) PR to implement JR telemetry.

rotorflight (approved):

PR #227, opened 2025-01-21
Fix battery and cell count detection

rotorflight:


rotorflight:
This is supposed to fix the issue with battery detection, when the battery is not detected properly and the cell count remains at zero. The cause is likely the max "noise" level allowed in the voltage measurement. It's set to 200mV at the moment, and is likely too tight for ESCs that are measuring the voltage with poor precision or lots of noise.

PR #232, opened 2025-01-29
Setpoint boost

gongtao0607:
This PR is base on previous experimental setpoint boost, with enhancement of separated cutoff frequencies.
I didn't add MSP so this will be cli secret sauce for now.

@etocii any suggestion on the default values?


etocii:
The pros seems to like:

setpoint_boost = 3500,3500,9000
setpoint_booset_cutoff = 15,15,90

Personally I am more confident of the cyclic boost numbers. I was initially trying to make tail stop stronger. While it worked on the M7 very well, I found it not to change much on other machines. The 9000 gain and 90 cutoff is still a safe number for people to try however.

gongtao0607:
[f943a6f](https://github.com/rotorflight/rotorflight-firmware/pull/232/commits/f943a6f4e56ddc296e75aa9cf07abd44b3842967)
Is the setpoint boost.
[32b2186](https://github.com/rotorflight/rotorflight-firmware/pull/232/commits/32b21866be63270ad6c328a65101b3799c66356d)
Is the unit test, it also includes some minor fixes:
    * Added `common/utils.h` for `BIT()`
    * Renamed `pid_t` because it conflicts with process id.

src/main/cli/settings.c
```diff
@@ -983,6 +983,9 @@ const clivalue_t valueTable[] = {
 
     { "cyclic_ring",                VAR_UINT8  | PROFILE_RATE_VALUE, .config.minmaxUnsigned = { 0, 100 }, PG_CONTROL_RATE_PROFILES, offsetof(controlRateConfig_t, cyclic_ring) },
 
+    { "setpoint_boost",             VAR_UINT8  | PROFILE_RATE_VALUE | MODE_ARRAY, .config.array.length = 4, PG_CONTROL_RATE_PROFILES, offsetof(controlRateConfig_t, setpoint_boost) },
```
pmattila, line 986: use `setpoint_boost_gain`

src/main/flight/setpoint.c
```diff
@@ -200,6 +208,10 @@ void setpointUpdate(void)
 
         SP = sp.setpoint[axis] = applyRatesCurve(axis, SP);
         DEBUG_AXIS(SETPOINT, axis, 4, SP);
+
+        // Apply boost
+        SP = sp.setpoint[axis] +=
+            difFilterApply(&sp.boostFilter[axis], SP) * sp.boostGain[axis];
```
pmattila, line 214: After this line, we should really have `DEBUG_AXIS(SETPOINT, axis, 5, SP);`
Need to change current use of 5 and 6 to 6 and 7.

src/main/flight/setpoint.c
```diff
@@ -200,6 +208,10 @@ void setpointUpdate(void)
 
         SP = sp.setpoint[axis] = applyRatesCurve(axis, SP);
         DEBUG_AXIS(SETPOINT, axis, 4, SP);
+
```
pmattila, line 211: Are you confident that the boost should be done after the rate curve application?

pmattila (changes requested):

pmattila:
Need to add MSP

src/main/flight/setpoint.c
gongtao0607: Why not just 7? Doesn't seem very manageable to change the existing ones. Also, we will soon add another for the dynamic deadzone.

src/main/flight/setpoint.c
gongtao0607: Since it's "Setpoint boost", I feel it's reasonable after the rate curve. Otherwise it would be more like "stick boost"? @etocii to justify.

src/main/cli/settings.c
gongtao0607: Done

gongtao0607:
Please take a look at [881f3d0](https://github.com/rotorflight/rotorflight-firmware/pull/232/commits/881f3d0c8109039c34d8d68f7eebdd75df7362ae) and [2ba2df7](https://github.com/rotorflight/rotorflight-firmware/pull/232/commits/2ba2df7486419837e581b2f1209688e48857b359)

MSP is added to `RC_TUNING`, since this is where `currentControlRateProfile` gets assigned

pmattila (approved):

src/main/pg/rates.c
```diff
@@ -67,6 +67,14 @@ void pgResetFn_controlRateProfiles(controlRateConfig_t *controlRateConfig)
             .accel_limit[FD_PITCH] = 0,
             .accel_limit[FD_YAW] = 0,
             .accel_limit[FD_COLL] = 0,
+            .setpoint_boost_gain[FD_ROLL] = 35,
+            .setpoint_boost_gain[FD_PITCH] = 35,
+            .setpoint_boost_gain[FD_YAW] = 0,
+            .setpoint_boost_gain[FD_COLL] = 0,
+            .setpoint_boost_cutoff[FD_ROLL] = 15,
+            .setpoint_boost_cutoff[FD_PITCH] = 15,
+            .setpoint_boost_cutoff[FD_YAW] = 0,
```
pmattila, line 76: Please set the cutoffs to reasonable values, so that it is enough to just set the gain.
Less questions from users.

src/main/pg/rates.c
```diff
@@ -67,6 +67,14 @@ void pgResetFn_controlRateProfiles(controlRateConfig_t *controlRateConfig)
             .accel_limit[FD_PITCH] = 0,
             .accel_limit[FD_YAW] = 0,
             .accel_limit[FD_COLL] = 0,
+            .setpoint_boost_gain[FD_ROLL] = 35,
```
pmattila, line 70: I am not sure we want to enable this feature by default.
Lost of users are already considering RF too snappy.

pmattila (changes requested):
Add Adjustment functions.

src/main/pg/rates.c
etocii: I think yaw cutoff at 90hz is what people ends up with. I haven't tested collective boost yet. An estimate would be 15hz

src/main/pg/rates.c
gongtao0607: All changed to 0

src/main/pg/rates.c
gongtao0607: 15, 15, 90, 15 now

pmattila (approved):

PR #241, opened 2025-02-07
Dynamic yaw deadband

gongtao0607:
cli:

dynamic_deadband_gain 0-250
dynamic_deadband_cutoff 0-250

(100, 30) = roughly 15% max deadband (for a very quick yaw centering)


pmattila:
Cleaned up and updated in #259

PR #248, opened 2025-02-16
Add Stop gain on yaw I

etocii:
There are two problems this change tries to solve:
1. Tail whips in CCW piro more than CW. 
2. Tail does not stop strong enough.

With high enough yaw I gain the problems can be solved. However, I found that yaw I gain on CCW stop behaves too high while it's tuned perfect for CW stops.

Some initial investigations suggests that this increases the yaw I gain from about 180 to 350 across 3 machines I tested. Yaw whipping is greatly reduced and there is no over gain phenomenon. (was obvious during ticktock before)


pmattila:
Included in #293

PR #253, opened 2025-02-27
 ESC sensor: HW4 spike filtering

wschoech:
The problem with HW4 is that, when there's no load on the head, the current readings have spikes in it. Eg. the current is 0 for a long period of time and suddenly a spike to 4A happens and then drops back to 0. I don't know whether the current readings have spikes when there's load on the head. However, the spikee current readings are apparently a problem and we should try getting rid of it.

There are 2  'filters' implemented
1 = kalman filter
2 = last X readings > 0, and only then consider the current reading valid, until a new 0 reading is seen, then this starts over

Another idea might be running average over the last X readings?
In general, I think the filtering should only be applied if current readings are below a certain threshold as otherwise consumption might be completely broken. 

Some testing: 
1: Smooths everything out and spikes are gone. Don't know the impact on the actual readings, might completely screw consumption etc
2: Is very dumb and the spikes are obviously not gone, just the 0 to e.g. 4A spikes. Spikes from e.g. 4 to e.g. 7 are still there

You can change the filter method via CLI:
kalman filter: set esc_sensor_hw4_filter_mode = 1
dumb 0 filter: set esc_sensor_hw4_filter_mode = 2

For the dumb 0 filter you can adjust the threshold via CLI
set esc_sensor_hw4_valid_current_readings_threshold = 5 (0-254)

Note: This is a draft version and I opened it for others to take a look, maybe test, and get comments on it. Code is not clean etc. 
Note2: I've no clue about filtering, kalman filters, whatsoever. I've no idea if the implemented kalman filter is correct, or the parameters are chosen appropriately.


robthomson:
Suggest we consider closing this PR as work @NightHawk32  is doing to fix the protocol looks to solve this.

pmattila:
Fixing the protocol instead.

PR #254, opened 2025-03-01
Dynamic Deadband

etocii:
This is a working algorithm on dynamic deadband. Code is not very polished but it gets the job done.


pmattila:
Cleaned up and updated in #259

PR #257, opened 2025-03-04
ESC sensor: Change HW4 motor-off detection from pwm to throttle

wschoech:
It has been observed that the PWM value, as reported by the ESC, can be greater than zero while the motor remains inactive. This leads to inaccurate current readings, resulting in completely incorrect consumption values. The throttle, as reported by the ESC, has been 0 in such cases.

This PR changes the 'motor-off detection' logic by using throttle instead of PWM, which should hopefully resolve the issue.

To be tested and confirmed.


pmattila:
This is good enough to go into the master. We will continue debugging the actual issue.

pmattila (approved):

PR #259, opened 2025-03-08
Add yaw dynamic deadband

pmattila:
Needs testing and tweaking.


src/main/flight/setpoint.c
```diff
@@ -203,16 +234,28 @@ void setpointUpdate(void)
         SP = filterApply(&sp.smoothingFilter[axis], SP);
         DEBUG_AXIS(SETPOINT, axis, 2, SP * 1000);
 
+        if (axis == FD_YAW) {
+            sp.yawDynamicDeadband = getYawDynamicDeadband(SP);
+            const float setpoint = fapplyDeadband(SP, sp.yawDynamicDeadband);
+            const float range = 1.0f - sp.yawDynamicDeadband;
```
etocii, line 240: We could also add a high deadband here by simply subtract a high deadband value. I.e.
```
Range = 1.0f - lowDynamicDeadband - highDynamicDeadband
```
And make sure range is above a small number

pmattila:
Need to add MSP support and Adjustment Functions. Then it's probably ready.

pmattila:
Only thing left is to decide if we need a parameter for the differentiator cutoff, or if we can use a constant like 20Hz.

PR #260, opened 2025-03-08
Ability to adjust BMI088 filters via CLI

NightHawk32:


pmattila (commented):
I think we can leave out the 400Hz option. Otherwise somebody may actually use it...

src/main/drivers/accgyro/gyro_sync.c
```diff
@@ -73,7 +74,19 @@ void gyroSetSampleRate(gyroDev_t *gyro)
             gyro->gyroRateKHz = GYRO_RATE_2000_Hz;
             gyroSampleRateHz = 2000;
             accSampleRateHz = 800;
-            break;
+            switch(gyroConfig()->gyro_hardware_lpf) {
```
pmattila, line 77: It's a good programming practise to handle all enum values in a switch-case.

pmattila (changes requested):

src/main/drivers/accgyro/accgyro_spi_bmi088.c
```diff
@@ -138,7 +138,31 @@ void bmi088SpiGyroInit(gyroDev_t *gyro)
...
 
     spiWriteReg(dev, BMI088_REG_GYRO_RANGE, BMI088_G_RANGE_2000DPS);
-    spiWriteReg(dev, BMI088_REG_GYRO_BANDWIDTH, BMI088_G_BANDWIDTH_230HZ);
+
+    uint8_t filtConf = 0;
+
+    switch(gyroConfig()->gyro_hardware_lpf) {
+        case GYRO_HARDWARE_LPF_NORMAL:
+            //ODR: 2Khz
+            filtConf = BMI088_G_BANDWIDTH_532HZ;
```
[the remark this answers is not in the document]
pmattila: Please change to:
NORMAL 230Hz
OPTION_1 532Hz
OPTION_2 Hmmm.... there is nothing higher... maybe use 532Hz
EXPERIMENTAL 116Hz

Then it would be in-line with the idea of higher "options"

src/main/drivers/accgyro/gyro_sync.c
NightHawk32: done!

pmattila (approved):

pmattila (changes requested):
We need to find out why this is not working.

NightHawk32:
> We need to find out why this is not working.

Works fine with my hardware. Got some additional hardware on the way. Once I have that I can further investigate.

pmattila (approved):

PR #263, opened 2025-03-16
Modify the temp2 data source of flyrotor ESC.

Jiki-Mo:
The temp2 data source of the flyrotor ESC is changed to the motor temperature.


Jiki-Mo:
This PR has been tested and achieved the expected results.

pmattila (approved):

PR #269, opened 2025-03-20
Fix rescue controls in hover state

pmattila:
Very hasty first version. UNTESTED!


pmattila:
I have now tested this _on the bench_
"Seems to work correctly"
Proceed with caution.

plesogor:
tested this IRL, no-flip, angle mode controls are working correctly inverted and upright. collective direction is also fine.  tested flip rescue as well and no issues here either.

PR #272, opened 2025-03-28
HW V4 telemetry optimizations

NightHawk32:
- should fix the current glitches to 0
- new parameter to set the threshold for valid current measurements (esc_sensor --> hw4_validCurrentThreshold)
- frames that should exist according to the HW spec should now also be read (mainly Type 1 in HW Spec)
- Experimental: dynamic calibration of the current offset, since its value differs for every ESC and can change over temperature. (can be overridden and manually set via CLI)
- gain is now calculated differently (https://github.com/dgatf/msrc/blob/master/board/project/sensor/esc_hw4.c) the developers took a closer look at the circuit and achieved better results. I opened a HW80 ESC and checked part of it.
- gain and offset can be overridden via CLI (as in the previous implementation (hw4_current_offset, hw4_current_gain)


pmattila:
Please keep in mind that the low level driver is not supposed to filter the data. Removing wrong samples is fine / preferred, but smoothing is not.
The next layer, the battery module does the filtering.

NightHawk32:
Sure, there is no filtering going on but rather a rejection of strange samples.

robthomson:
My tests of this show that it is better.  but really we need some wider testing!

NightHawk32:
@pmattila reverted all the white space changes. For some reason, my VS Code auto formatted the code and didn't even show me the changes.

robthomson:
This one seems to be stuck in the list.  Are we waiting on anything?

pmattila:
Comparing to my notes from the time I wrote the driver, and the data dumps I received for various V4 ESCs...
This code is not working correctly.
Also, the protocol description from HW is incorrect. Their ESCs do not behave as described in it.

PR #275, opened 2025-03-30
Changes made to the XBus Mode A allowing for varying packet len

TheNige069:
Changes made to the XBus Mode A protocol to allow for varying packet length depending on TX channels. Some minor improvements to help robustness of code

### DIFF
--- a/src/main/rx/xbus.c
+++ b/src/main/rx/xbus.c
@@ -104,6 +104,7 @@
 #define XBUS_MODEA_CONVERT_TO_USEC(V) (800 + ((V * 1400) >> 16))
 // As Mode A has a difference frame time, lets set it in a global
 static timeDelta_t xBusMaxFrameTime;
+static timeUs_t xBusTimeLast = 0;
 
 static bool xBusFrameReceived = false;
 static bool xBusDataIncoming = false;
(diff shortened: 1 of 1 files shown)


src/main/rx/xbus.c
```diff
@@ -349,15 +335,26 @@ static void xBusDataReceive(uint16_t c, void *data)
...
+    frameStatus = RX_FRAME_COMPLETE;
 
-    return RX_FRAME_COMPLETE;
+    if (rxRuntimeState->serialrxProvider == SERIALRX_XBUS_MODE_A) {
+        if (xBusFrame[2] == 0x00 && xBusFrame[3] == 0x80) {
+            frameStatus = RX_FRAME_COMPLETE | RX_FRAME_FAILSAFE;
+        }
+    }
+    
+    if (!(frameStatus & (RX_FRAME_FAILSAFE | RX_FRAME_DROPPED))) {
```
pmattila, line 353: Just checking... This evaluates true if _neither_ FAILSAFE _nor_ DROPPED is set. Correct?

TheNige069: That is correct. Based on the logic, if frameStatus equals RX_FRAME_COMPLETE, then it will update the lastRcFrameTimeUs value. If it is in failsafe or framelost, then the value doesn't get updated.

pmattila (approved):

Issue #302, opened 2025-05-05
Telemetry sensor BEC1  Voltage  - Invalid reading


jimmy6616:
Ethos 1.6.3, RF RC-3, 

Incorrect readings of alternate Bec voltages as below, affecting F.bus and CRSF

Ethos  - BEC Voltage - Sensor ID 0901 reading correctly
              BEC1 Voltage - Sensor ID 0219 Invalid reading

CRSF - BEC Voltage - Sensor ID 1081 reading correctly
             ESC1 Bec Voltage - Sensor ID 1049 Invalid reading

In both cases the configurator is showing a correct BEC voltage

![Image](https://github.com/user-attachments/assets/fe0fd01e-8942-4828-969e-1a4dba05e7e2)
![Image](https://github.com/user-attachments/assets/8929fd8f-5842-4159-b8cf-64a20edb23ed)
![Image](https://github.com/user-attachments/assets/a2d36265-6ff4-4861-b500-7cf42d0bc143)

rburrow87:
Which ESC is this?

jimmy6616:
Hobbywing v5

robthomson:
@rburrow87 are you going to check it out?

jimmy6616:
Its the same with BLHeli32 on an Escape32 ESC

robthomson:
It will be the sensor code on fbl not using correct sensor.  @pmattila ...   I can tackle this but am not familiar with code on this front.  Maybe quicker for you?

robthomson:
Potential fix here:

https://github.com/rotorflight/rotorflight-firmware/actions/runs/14881872445

robthomson:
I have not seen same issue my end

jimmy6616:
Rob ive created a new base config on the Nexus, this time with the X20, same result, im using the default RC4 software, no PR's
    On Sunday 11 May 2025 at 12:20:35 BST, Rob Thomson ***@***.***> wrote:  
 
 robthomson left a comment (rotorflight/rotorflight-firmware#302)
I have not seen same issue my end

jimmy6616:
I need to check, I dont believe it has, will update tomorrow

PR #303, opened 2025-05-07
Invalid reading for bec1/bec2 voltages

robthomson:
This PR should fix this issue:

https://github.com/rotorflight/rotorflight-firmware/issues/302


jimmy6616:
will test today Rob

robthomson:
> The purpose of the ESC BEC telemetry items is to access BEC voltage as reported by the ESC. Making it not do that defeats the purpose of the sensor.

Interesting.   This is not particually clear to end users who will assume this is just the same - but more specific.

What we could do is rather than doing the PR in this way.. I could just alter the scaling for the sensor.

robthomson:
Further tests have shown this is not the source issue.  Need to investigate more.

PR #306, opened 2025-05-09
first draft for am32 fwd prog

NightHawk32:


heliflieger:
What I the state of this pull request?

NightHawk32:
Pretty much outdated. I will check and update this on Monday. We need to update suite as well. The memory layout of the current ESC FW has also changed by now.

NightHawk32:
@heliflieger I have updated this PR. Reading from the ESC works; writing is not tested, as of now. There are currently some timing issues on suite side.

heliflieger:
Yes Lukas works on it

src/main/sensors/esc_sensor.h
```diff
@@ -59,6 +59,13 @@ uint32_t getEscSensorRPM(uint8_t motorNumber);
 escSensorData_t *getEscSensorData(uint8_t motorNumber);
 
 uint8_t escGetParamBufferLength(void);
+/**
+ * Select 4-way interface ESC by id.
+ *
+ * @param id ESC identifier (0..MAX_SUPPORTED_MOTORS-1). Use 0xFF to deselect the 4-way interface.
+ * @return 0 on success, non-zero on error (e.g. invalid id).
+ */
+uint8_t escSelect4WIfById(uint8_t id);
```
pmattila, line 62: We are not using doxygen, so these comments are a bit pointless

src/main/io/serial_4way.c
```diff
@@ -156,13 +154,18 @@ uint8_t esc4wayInit(void)
...
 
-void esc4wayRelease(void)
+void esc4wayDeinit(void)
 {
-    while (escCount > 0) {
-        escCount--;
-        IOConfigGPIO(escHardware[escCount].io, IOCFG_AF_PP);
-        setEscLo(escCount);
+    uint8_t cnt = escCount;
+    while (cnt > 0) {
```
pmattila: Use a for loop whenever reasonable. It's most common and easier to read.

src/main/io/serial_4way.c
```diff
@@ -348,6 +351,9 @@ static uint8_t Connect(uint8_32_u *pDeviceInfo)
             CurrentInterfaceMode = imSK;
             return 1;
         } else {
+            uint32_t startTime = micros();
+            while(micros() < startTime + 10000) {
```
pmattila: I think there is a sleep function

pmattila (changes requested):

src/main/target/common_pre.h
```diff
@@ -397,6 +397,8 @@ extern uint8_t _dmaram_end__;
 #define USE_CUSTOM_BOX_NAMES
 #define USE_RX_LINK_UPLINK_POWER
 #define USE_CRSF_V3
+#define USE_SERIAL_4WAY_BLHELI_BOOTLOADER
+#define USE_SERIAL_4WAY_SK_BOOTLOADER
```
pmattila, line 401: The code does not compile if either of these is undefined.

Rule: if the code uses ifdefs, it must at least build with all combinations.

src/main/io/serial_4way.c
```diff
@@ -45,9 +45,7 @@
 #ifdef USE_SERIAL_4WAY_BLHELI_BOOTLOADER
 #include "io/serial_4way_avrootloader.h"
 #endif
-#if defined(USE_SERIAL_4WAY_SK_BOOTLOADER)
 #include "io/serial_4way_stk500v2.h"
-#endif
```
pmattila, line 50: Check if removing this #if is the right thing to do

PR #310, opened 2025-05-26
Spektrum SRXL2 Rx, force half duplex in a way so RFC can see it.

AndersHoglund:
Currently SRXL2 half duplex serial is forced by the driver but RFC does not show it.


pmattila:
I can't see where the driver is forcing half duplex?

pmattila (approved):

PR #316, opened 2025-06-22
Change default for  cbat_alert_percent  to 35%

robthomson:
This PR updates the default value for

set cbat_alert_percent

from 10  to 35

With the use of this setting becoming a used feature in suite; a more usefull and safe min pack threshold should be used.


pmattila:
It's used for determining the battery state, which can be indicated in OSD, LED colours, and beeper.

PR #317, opened 2025-06-22
feat: added flight statistics to MSP_FLIGHT_STATS

egonl:
The flight statistics were only accessible through CLI. This PR adds support for MSP by extending MSP_PILOT_CONFIG with those stats, so they can be used by e.g. [Lua](https://github.com/rotorflight/rotorflight-lua-scripts/pull/108):

image
 
Also added modelFlags to the pilotConfig_t structure, for model specific features, e.g. 'set name of model on the radio', or 'announce the battery capacity left'.


robthomson:
Thats really usefull extra stuff.

Will use it as soon as it merges!

robthomson:
I do personally wonder if the pilot config msp is the right place for this?

Surely it should be a separat msp call for stats?

robthomson:
It's no big issue either way - Juet not an obvious place to look when searching for msp calls / functions

pmattila (approved):

PR #334, opened 2025-08-27
Fix JR Xbus Mode-A jittering and random motor cuts

TheNige069:
The code changes fix the following:

- fixed the jittering that was being experienced in the Rotorflight Configurator when displaying the channels
- Above fix didn't work for transmitters with less than 16 channels. Subsequent commits fixed this issue
- Re-wrote some of the code to improve processing speed
- Tidied up and removed some comments
- Fixed the random "motor cuts" that I was experiencing. 30+ flights without a motor cut, whereas before these fixes, I was getting about one motor cut every few flights

Note: I reverted a change I made to workflows/push.yml. I didn't want this change included in this PR as not required.


TheNige069:
As of yesterday (Sunday), I have 40 flights on this firmware revision without any issues. As part of yesterday's flights, I switched between my JR Elite and JR XG14 and flew one of my helis with each radio with no issues.
Please review this PR.
Thank you.

src/main/rx/xbus.c
```diff
@@ -160,45 +160,47 @@ static uint8_t xBusRj01CRC8(uint8_t inData, uint8_t seed)
...
+    // CRC check (per JR XBUS Mode A spec)
+    if (crc8_dallas((uint8_t*)xBusFrame, xBusFrame[1] + 2) != xBusFrame[xBusFrame[1] + 2]) {
+        return; // CRC failed
+    }
+
+    // Need to do a check on bytes 2 and 3. 
+    // When failsafe, byte 2 goes from >0 to 0 and byte 3 goes from 0 to a value that appears to be dependant on the TX
+    // XG14 goes 0x1c 0x00 to 0x00 0x80
+    // Elite goes 0x17 0x00 to 0x00 0x80
+   if (xBusFrame[2] == 0x00 && xBusFrame[3] == 0x80) {
```
pmattila: Please check if this line is correct. It's not an inverse of the original.

TheNige069: It is the inverse. The original was a not equal (!=), where as in the line above (172), we are now testing for equality. The logic is still the same. If Frame2 is 0x0 and Frame3 is 0x80, then we have a fail safe condition and as such, return

pmattila (approved):

src/main/rx/xbus.c
pmattila: Look more closely. You missed the && to || change.
Unless... the orginal was wrong. I was not able to tell.

PR #344, opened 2025-11-04
Fix MSPv2 over SmartPort: repeated START frames after header-only first chunk

robthomson:
### Summary
This PR fixes an issue where MSPv2 telemetry over **SmartPort** would continuously resend the first “START” frame and never send continuation frames, preventing MSPv2 replies from completing.

### Root Cause
SmartPort limits each telemetry frame to 6 bytes.  
When `sendMspReply()` sends the first MSPv2 reply frame, that frame often contains **only the 6-byte header** (no payload bytes).  

The function previously used:
```c
if (responsePacket.buf.ptr == responseBuffer)
```
to detect whether the current chunk is the first frame.  
Because the header-only case doesn’t advance the buffer pointer, subsequent calls still satisfy the condition — every chunk is treated as “first”, so the firmware keeps setting the `MSP_STATUS_START_MASK` bit.

### Fix
A minimal fix introduces a small static flag (`headerSent`) to track whether the header has already been emitted:

```c
static bool headerSent = false;

if (!headerSent) {
    // First frame: add START flag + header
    ...
    headerSent = true;
} else {
    // Continuation: no START flag
    ...
}

...

// Reset flag after final chunk
headerSent = false;
```

This ensures only the first frame of a response has the START flag, even if it carried no payload bytes.

### Verification
| Transport | MSP Version | Result |
|------------|--------------|--------|
| SmartPort  | MSPv1 | ✅ Unchanged - checked |
| SmartPort  | MSPv2 | ✅ Fixed – now sends continuation frames and completes replies |
| CRSF       | MSPv1 | ✅ Unchanged - checked|
| CRSF       | MSPv2 | ✅ Unchanged - checked |

Logs now show:
```
RXv2 start … continuation expected
RXv2 continuation … collected=3/3
RXv2 complete
```

### Impact
- **Fixed:** MSPv2 over SmartPort telemetry now functions correctly  
- **No change:** MSPv1 or CRSF behavior  
- **Low risk:** The fix only affects chunk sequencing logic

### Tags
`telemetry` • `smartport` • `mspv2` • `bugfix`

---

### DIFF
--- a/src/main/telemetry/msp_shared.c
+++ b/src/main/telemetry/msp_shared.c
@@ -325,7 +325,7 @@ bool sendMspReply(const uint8_t payloadSizeMax, mspResponseFnPtr responseFn)
     sbufSwitchToReader(&responsePacket.buf, responseBuffer);// for CRC calculation
 
     responseFn(payloadArray, payloadBuf->ptr - payloadArray);
-    headerSent = false;               // reset for the next response
+    headerSent = false; // reset for the next response
     return false;
 }
 


src/main/telemetry/msp_shared.c
```diff
@@ -323,6 +325,7 @@ bool sendMspReply(const uint8_t payloadSizeMax, mspResponseFnPtr responseFn)
     sbufSwitchToReader(&responsePacket.buf, responseBuffer);// for CRC calculation
 
     responseFn(payloadArray, payloadBuf->ptr - payloadArray);
+    headerSent = false;               // <-- added: reset for the next response
```
bsongis: seems ok as a fix
I would just change the comment: no `<-- added:`

pmattila: I agree 100%. 
If that kind of comment was necessary, every line should have it, as they were adder at some point!

PR #351, opened 2025-12-06
ZTW Fwd programing support

robthomson:
This PR relies on the preceeding PR

https://github.com/rotorflight/rotorflight-firmware/pull/347

Support is added to include ZTW fwd programing and configuration.   

**Tested and working in legacy mode without a firmware that supports fwd programing**
ztw-legacy mode works

Tested with old firmware and new firmwared.  All works as expected.

### DIFF
--- a/src/main/sensors/esc_sensor.c
+++ b/src/main/sensors/esc_sensor.c
@@ -3496,7 +3496,7 @@ static int8_t xdfly_accept(uint16_t c)
     return 0;
 }
 
-static serialReceiveCallbackPtr xdflySensorInitCommon(uint8_t sig)
+static serialReceiveCallbackPtr xdflySensorInit(uint8_t sig)
 {
     rrfsmCrank  = xdfly_crank_unc_setup;
     rrfsmDecode = xdfly_decode;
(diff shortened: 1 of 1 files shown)


src/main/sensors/esc_sensor.c
```diff
@@ -3611,23 +3489,19 @@ static int8_t xdfly_accept(uint16_t c)
...
 
-static serialReceiveCallbackPtr ompAsXdflySensorInit(void)
-{
-    serialReceiveCallbackPtr cb = xdflySensorInit();
-    escSig = ESC_SIG_OMP;   // 0xD0 for OMP header
-    return cb;
-}
+static serialReceiveCallbackPtr xdflySensorInit(void)      { return xdflySensorInitCommon(ESC_SIG_XDFLY); }
+static serialReceiveCallbackPtr ompAsXdflySensorInit(void) { return xdflySensorInitCommon(ESC_SIG_OMP); }
+static serialReceiveCallbackPtr ztwAsXdflySensorInit(void) { return xdflySensorInitCommon(ESC_SIG_ZTW); }
```
pmattila: These are totally unnecessary. Please use the xdFly... function direction in the switch-case below

src/main/sensors/esc_sensor.c
```diff
@@ -3336,7 +3214,7 @@ static enum xdfly_setup_status xdfly_setup_status = XDFLY_INIT;
 
 static inline uint8_t xdfly_sync_header(void)
 {
-    return (escSig == ESC_SIG_OMP) ? OMPHOBBY_SYNC : XDFLY_SYNC;
+    return (escSig == ESC_SIG_OMP) ? OMPHOBBY_SYNC : (escSig == ESC_SIG_ZTW) ? ZTW_SYNC : XDFLY_SYNC;
```
pmattila: It's bad coding style to nest multiple ?:
Please use swtich-case instead

pmattila (changes requested):
A few tiny fixed needed.

PR #354, opened 2025-12-16
Spektrum ESC telemetry output to Tx added. 

AndersHoglund:
If some ESC telemetry sensor is enabled, it would be nice to have it shown on Tx as ESC telemetry. Even though most or all data elements already show up on other tlm pages, a collected view would be nice. 
Tested with hardcoded test data (screenshot below), and SDfly Hpro65 ESC, iX20, Nexus, SPM4650.  Spektrum has a problem in their end, shows BEC Amps (A) not mA.
 
Screenshot_2025-12-16_125621

Mininal radio extra bandwith wasted.
Screenshot_2025-12-16_130803
/A

### DIFF
--- a/src/main/telemetry/srxl.c
+++ b/src/main/telemetry/srxl.c
@@ -456,7 +456,9 @@ bool srxlFrameFlightPackCurrent(sbuf_t *dst, timeUs_t currentTimeUs)
     return false;
 }
 
-//  ****** ESC Telem*try frame ******
+#ifdef USE_ESC_SENSOR_TELEMETRY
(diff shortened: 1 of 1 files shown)


QuPsi:
@AndersHoglund I just went through your code and noticed that you only populated the ESC frame. It's possible to get mAh consumed as well, but it uses header 0x34. You can see the structure here.

https://github.com/SpektrumRC/SpektrumDocumentation/blob/master/Telemetry/spektrumTelemetrySensors.h#L80

PR #355, opened 2025-12-17
FBUS Master Out

NightHawk32:
### Ready for testing and review

- added some new parameters to configure FBUS
- Find my RS4 config attached:
[RTFL_cli_20251217_140628_FBUS.txt](https://github.com/user-attachments/files/24213746/RTFL_cli_20251217_140628_FBUS.txt)
- This has been tested for quite a while on my RS4
- Servo config was done via Radio: front CH1 6800, left CH3 6802, right CH2, 6802, tail, CH4 6803 --> all servos connected to FBUS out on Vantac 007 via servo splitter
Screenshot_20251217_144710_AliExpress
original_9ddbeac4-13b0-487a-8fa1-c3f83c997d5e_20250611_151536

HW Setup:
- connect the FBUS wire to a free UART TX on your flight controller
- you may need *fbus_master_pinswap ON*

Config
**Serial port function value**: `524288` (this is the bitmask for `FUNCTION_FBUS_MASTER`)
**Note**: The baud rate parameters in the `serial` command are ignored for FBUS Master since it always uses 460800 baud. However, you still need to set the function bitmask to 524288.

Parameters
```
# Source types:
# 0 = NONE (disabled)
# 1 = RX (receiver input channels)
# 2 = MIXER (mixer rule outputs)
# 3 = SERVO (servo outputs)
# 4 = MOTOR (motor outputs)

# Example: Map first 8 channels to RX inputs
set fbus_master_source_type = 3,3,3,3,1,1,1,1,1,1,1,1,1,1,1,1
set fbus_master_source_index = 1,2,0,3,4,5,6,7,8,9,10,11,12,13,14,15
```
```
# For standard RX channels (1000-2000 µs)
set fbus_master_source_range_high = 2000,2000,2000,2000,2000,2000,2000,2000,2000,2000,2000,2000,2000,2000,2000,2000
set fbus_master_source_range_low = 1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000,1000

# For mixer outputs (-1000 to +1000, representing -1.0 to +1.0)
# set fbus_master_source_range_low = -1000,-1000,...
# set fbus_master_source_range_high = 1000,1000,...
```
```
# Set frame rate to 500 Hz (recommended for most receivers)
set fbus_master_frame_rate = 500

# Valid range: 25-550 Hz
```

Thanks to FrSky for helping me out with Xact Servos!
3x M5252H, 1x MD5301H

### DIFF
--- a/src/main/msp/msp.c
+++ b/src/main/msp/msp.c
@@ -2104,6 +2104,24 @@ static mspResult_e mspFcProcessOutCommandWithArg(mspDescriptor_t srcDesc, int16_
             serializeBoxReply(dst, page, &serializeBoxPermanentIdFn);
         }
         break;
+    case MSP_GET_SBUS_OUTPUT_CONFIG:
+        {
+            const int rem = sbufBytesRemaining(src);
+            if (rem != 1) {
+                return MSP_RESULT_ERROR;
+            }
+
+            const uint8_t index = sbufReadU8(src);
+            if (index >= SBUS_OUT_CHANNELS) {
+                return MSP_RESULT_ERROR;
+            }
+
+            sbufWriteU8(dst, sbusOutConfigMutable()->sourceType[index]);
+            sbufWriteU8(dst, sbusOutConfigMutable()->sourceIndex[index]);
+            sbufWriteS16(dst, sbusOutConfigMutable()->sourceRangeLow[index]);
+            sbufWriteS16(dst, sbusOutConfigMutable()->sourceRangeHigh[index]);
+        }
+        break;     
     case MSP_GET_MIXER_INPUT:
         {
             const int rem = sbufBytesRemaining(src);
--- a/src/main/msp/msp_protocol.h
+++ b/src/main/msp/msp_protocol.h
@@ -200,6 +200,7 @@
 #define MSP_RPM_FILTER_V2                    154
 #define MSP_SET_RPM_FILTER_V2                155
 #define MSP_GET_ADJUSTMENT_RANGE             156
+#define MSP_GET_SBUS_OUTPUT_CONFIG           157
 
 #define MSP_EXPERIMENTAL                     158
 #define MSP_SET_EXPERIMENTAL                 159
--- a/Changes.md
+++ b/Changes.md
@@ -64,6 +64,21 @@ New MSP command (162) to configure individual FBUS Master channel settings.
 
 New MSP command (163) to retrieve individual FBUS Master channel configuration.
 
+### MSP_GET_SBUS_OUTPUT_CONFIG
+
+Allow querying a single sbus servo via msp (#372)
+
+### MSP_GET_MIXER_INPUT 
+
+Add msp call to allow retrieving a single mixer line at a time (#361)
+
+### MSP_GET_ADJUSTMENT_RANGE
+
+Add msp call to allow retrieving a single adjustment line at a time (#362)
+
+### MSP_SET_SERVO_CENTER 
+
+Add msp call to set just the servo center point (#366)
 
 ## CLI Changes
 


src/main/msp/msp_protocol.h
```diff
@@ -199,6 +199,7 @@
 #define MSP_SET_SBUS_OUTPUT_CONFIG           153
 #define MSP_RPM_FILTER_V2                    154
 #define MSP_SET_RPM_FILTER_V2                155
+#define MSP_SET_FBUS_MASTER_CONFIG           156
```
robthomson: Suggest we do 156 Read and 157 Write

src/main/msp/msp.c
```diff
@@ -1731,6 +1732,17 @@ static bool mspProcessOutCommand(int16_t cmdMSP, sbuf_t *dst)
         break;
 #endif
 
+#ifdef USE_FBUS_MASTER
```
robthomson, line 1735: Pretty sure this should be the SET msp.. as the current call implies its a 'read'

src/main/msp/msp_protocol.h
robthomson: Resolved

src/main/msp/msp.c
robthomson: Resolved

NightHawk32:
Todo: remove the MUX code and put it into a separate PR. This is not required for the initial version

NightHawk32:
Found a bug, FBUS frame rate can't be set to >255Hz due to the value being uin8t. ToDo: fix that

robthomson (commented):
I have tested this on a 12 hour soak test.

Zero issue with servo control.   LGTM

src/main/common/utils.h
```diff
@@ -69,6 +69,7 @@
 
 #define BIT(x) (1 << (x))
 #define BITLL(x) (1ULL << (x))
+#define GET_BIT(value, bit) ((value >> bit) & 1)
```
pmattila: Please move this into fbus_master.c, as it is quite specific to the way you use it.

src/main/drivers/fbus_master.c
```diff
@@ -0,0 +1,309 @@
...
+    frame->c16.rssi = 100; //ToDo
+
+    uint8_t crc = frskyCheckSum((uint8_t *)&(frame->c16.type), FBUS_MASTER_CONTROL_FRAME_PAYLOAD_SIZE);
+    frame->c16.crc = crc;
+
+    switch (fbusMasterPayloadState)
+    {
+    case FBUS_MASTER_TELEMETRY:
+
+         memset(&frame->downlink, 0, sizeof(fbusMasterDownlink_t));
```
pmattila: indentation

src/main/drivers/sdio_f7xx.c
```diff
@@ -1366,7 +1366,7 @@ SD_Error_t SD_GetCardStatus(SD_CardStatus_t* pCardStatus)
 static SD_Error_t SD_PowerON(void)
 {
     SD_Error_t ErrorState;
-    uint32_t   Response;
+    uint32_t   Response = 0;
```
pmattila: This should not be in this PR

src/main/drivers/fbus_master.c
```diff
@@ -0,0 +1,309 @@
...
+        readIngoreBytes--;
+        return;
+    }
+
+    buffer[readBytes++] = c;
+
+    if (readBytes >= FBUS_MASTER_BUFFER_SIZE) {
+        //sync error
+        readBytes = 0;
+    }else{
```
pmattila: Please follow a consistent use of spaces around bracket, curly braces etc.
This should be `} else {`

src/main/drivers/fbus_master.c
```diff
@@ -0,0 +1,309 @@
...
+        if(readBytes >= 10 && buffer[0] == FBUS_DOWNLINK_PAYLOAD_SIZE /*&& buffer[1] <= FBUS_MAX_PHYS_ID*/){
```
pmattila: Consistent use of space. Should be `if (`

src/main/drivers/fbus_master.c
```diff
@@ -0,0 +1,309 @@
...
+    }
+    fbusMasterPrepareFrame(&frame, channels);
+
+    // serial output
+    serialWriteBuf(fbusMasterPort, (const uint8_t *)&frame, sizeof(frame));
+    readIngoreBytes = sizeof(frame);
+    readBytes = 0;
+}
+
+bool fbusMasterIsEnabled()
```
pmattila: Please use C-stype function definition. This should be `(void)`

src/main/drivers/fbus_master.c
```diff
@@ -0,0 +1,309 @@
...
+    readIngoreBytes = sizeof(frame);
+    readBytes = 0;
+}
+
+bool fbusMasterIsEnabled() 
+{ 
+    return fbusMasterPort != NULL;
+}
+
+void fbusMasterInit()
```
pmattila: `(void)`

src/main/drivers/fbus_master.h
```diff
@@ -0,0 +1,108 @@
...
+    fbusMasterDownlink_t downlink;
+} __attribute__((__packed__)) fbusMasterFrame_t;
+
+STATIC_ASSERT(sizeof(fbusMasterControl16_t) == FBUS_MASTER_CONTROL_FRAME_SIZE,
+              fbus_master_frame_size_mismatch);
+
+// Routine function called by the scheduler or timer
+void fbusMasterUpdate(timeUs_t currentTimeUs);
+
+bool fbusMasterIsEnabled();
```
pmattila: `(void)`

src/main/fc/tasks.c
```diff
@@ -421,6 +423,11 @@ task_attribute_t task_attributes[TASK_COUNT] = {
     // 25Hz is the initial period. The actual period will be loaded from config.
     [TASK_SBUS_OUTPUT] = DEFINE_TASK("SBUS_OUTPUT", NULL, NULL, sbusOutUpdate, TASK_PERIOD_HZ(25), TASK_PRIORITY_MEDIUM),
 #endif
+
+#ifdef USE_FBUS_MASTER
+    // 25Hz is the initial period. The actual period will be loaded from config.
```
pmattila: Comment says 25Hz, code says 144Hz?

src/main/fc/init.c
```diff
@@ -696,6 +696,12 @@ void init(void)
     sbusOutInit();
 #endif
 
+
```
pmattila: Avoid excessive newlines. One in normal cases. Two if you need to make a statement that the following is widely different from the above.

pmattila (changes requested):

src/main/msp/msp_protocol.h
```diff
@@ -205,6 +205,9 @@
 #define MSP_SET_EXPERIMENTAL                 159
 
 #define MSP_UID                              160
+#define MSP_GET_FBUS_MASTER_CONFIG           161
```
pmattila, line 209: Okay this is nitpicking...
The "normal" get functions are usually not named with GET_ (... yes nitpicking)
Let's call this MSP_FBUS_MASTER_CONFIG.

But let's leave the GET_ and SET_ in the channel specific ones, to make a point.

pmattila (approved):

PR #356, opened 2025-12-18
Fix incomplete USE_ESC_SENSOR build deps.

AndersHoglund:
Building firmware with USE_ESC_SENSOR undefined fails. 
Might be of academic interest as we always have ESC sensor support included, but I think this build condition should either work or be removed.
/A

### DIFF
--- a/src/main/telemetry/sensors.c
+++ b/src/main/telemetry/sensors.c
@@ -205,10 +205,10 @@ int telemetrySensorValue(sensor_id_e id)
             return getCurrent(CURRENT_METER_ID_BUS);
         case TELEM_MCU_CURRENT:
             return getCurrent(CURRENT_METER_ID_MCU);
-        case TELEM_MCU_TEMP:
-            return getCoreTemperatureCelsius();
         case TELEM_TEMP:
             return millis();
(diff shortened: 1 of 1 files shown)


src/main/telemetry/sensors.c
```diff
@@ -202,15 +205,16 @@ int telemetrySensorValue(sensor_id_e id)
             return getCurrent(CURRENT_METER_ID_BUS);
         case TELEM_MCU_CURRENT:
             return getCurrent(CURRENT_METER_ID_MCU);
-
-        case TELEM_TEMP:
-            return millis();
         case TELEM_MCU_TEMP:
             return getCoreTemperatureCelsius();
+#ifdef USE_ESC_SENSOR
+        case TELEM_TEMP:
```
pmattila, line 208: TELEM_TEMP should not be within ifdef

AndersHoglund: Of course not.  Will fix.

pmattila: Please don't reorder the cases unnecessarily

pmattila (approved):

PR #358, opened 2025-12-29
SBUS Out Inversion

robthomson:
Added an cli option to allow the sbus out port to send an inverted/non inverted signal.

set sbus_out_inverted = ON | OFF

By default we use set sbus_out_inverted = ON  as this is wat 99% of systems will expect.  

In a few edge cases where devices do not have a hardware invertor; this is a usefull parameter to enable the fbl to invert the signal.


src/main/pg/sbus_output.c
```diff
@@ -36,6 +37,10 @@ void pgResetFn_sbusOutConfig(sbusOutConfig_t *config) {
     config->frameRate = 50;
 
     config->pinSwap = 0;
+
+    // Default to inverted S.Bus (normal for S.Bus receivers).
+    config->inverted = 1;
```
NightHawk32, line 40: replace this with SERIAL_INVERTED

src/main/drivers/sbus_output.c
```diff
@@ -167,7 +167,8 @@ void sbusOutInit()
 
     sbusOutPort = openSerialPort(
         portConfig->identifier, FUNCTION_SBUS_OUT, NULL, NULL, 100000, MODE_TX,
-        SERIAL_STOPBITS_2 | SERIAL_PARITY_EVEN | SERIAL_INVERTED |
+        SERIAL_STOPBITS_2 | SERIAL_PARITY_EVEN |
+            (sbusOutConfig()->inverted ? SERIAL_INVERTED : 0) |
```
NightHawk32: (sbusOutConfig()->inverted ? SERIAL_INVERTED : SERIAL_NOT_INVERTED)

robthomson: Done

src/main/pg/sbus_output.c
robthomson: This I think is better if we keep it to the number as it conforms with the other parameters not using a file thats included from elsewhere.

src/main/pg/sbus_output.c
```diff
@@ -23,8 +23,9 @@
 #ifdef USE_SBUS_OUTPUT
 
 // The config struct is quite large. A ResetFn is smaller than a ResetTemplate.
+// Version bumped due to added 'inverted' field.
```
pmattila: We don't put version specific comments in the code. It will clutter the code, as the comment is valid only for a moment. Nobody will remember to remove it later.

PR #360, opened 2025-12-29
BMP581 Driver

NightHawk32:
I2C and SPI implemented (I2C is tested)

- Power mode: NORMAL (continuous)
- Output Data Rate (ODR): 20 Hz (BMP581 ODR code set to BMP581_ODR_20_HZ)
- Pressure Oversampling (OSR_P): 32x (BMP581_OVERSAMPLING_32X)
- Temperature Oversampling (OSR_T): 128x (BMP581_OVERSAMPLING_128X)
- LPF in driver: disabled (use base firmware filtering)

Testing was done on RF dev board:
- Readings look reasonable
- I2C detection works as expected
- Temperature looks reasonable
- Pressure reading is alright (was compared against two reference sensors)


NightHawk32:
Update: SPI is now tested as well

src/main/drivers/barometer/barometer_bmp581.c
```diff
@@ -0,0 +1,424 @@
...
+static DMA_DATA_ZERO_INIT uint8_t sensor_data[BMP581_DATA_FRAME_SIZE];
+
+// Forward declarations
+static void bmp581StartUT(baroDev_t *baro);
+static bool bmp581GetUT(baroDev_t *baro);
+static bool bmp581ReadUT(baroDev_t *baro);
+static void bmp581StartUP(baroDev_t *baro);
+static bool bmp581GetUP(baroDev_t *baro);
+static bool bmp581ReadUP(baroDev_t *baro);
+STATIC_UNIT_TESTED void bmp581Calculate(int32_t *pressure, int32_t *temperature);
```
pmattila: Where is the unit test for this function?

src/main/drivers/barometer/barometer_bmp581.c
```diff
@@ -0,0 +1,424 @@
...
+    if (dev->bus->busType == BUS_TYPE_SPI) {
+        spiPreinitByIO(dev->busType_u.spi.csnPin);
+    }
+#else
+    UNUSED(dev);
+#endif
+}
+
+// Configure sensor mode (NORMAL or FORCED)
+void bmp581SetMode(const extDevice_t *dev, uint8_t powermode)
```
pmattila: Should be `static` ?

src/main/drivers/barometer/barometer_bmp581.c
```diff
@@ -0,0 +1,424 @@
...
+        IOHi(dev->busType_u.spi.csnPin);
+        spiSetClkDivisor(dev, spiCalculateDivider(BMP581_MAX_SPI_CLK_HZ));
+    }
+#else
+    UNUSED(dev);
+#endif
+}
+
+// Bus deinitialization
+void bmp581BusDeinit(const extDevice_t *dev)
```
pmattila: Should be `static` ?

src/main/drivers/barometer/barometer_bmp581.c
```diff
@@ -0,0 +1,424 @@
...
+    // TODO: Extract pressure and temperature from sensor_data
+    // This is a placeholder implementation
+    bmp581_up = (uint32_t)sensor_data[5] << 16 | (uint32_t)sensor_data[4] << 8 | sensor_data[3];
+    bmp581_ut = (uint32_t)sensor_data[2] << 16 | (uint32_t)sensor_data[1] << 8 | sensor_data[0];
+
+    return true;
+}
+
+// Calculate compensated pressure and temperature
+STATIC_UNIT_TESTED void bmp581Calculate(int32_t *pressure, int32_t *temperature)
```
pmattila: Yeah where is the unit test?

src/main/drivers/barometer/barometer_bmp581.c
```diff
@@ -0,0 +1,424 @@
...
+{
+    // BMP581 provides compensated values directly from the sensor
+    // Temperature: raw_data is 24-bit signed, divide by 2^16 (65536) to get °C
+    // Pressure: raw_data is 24-bit unsigned, divide by 2^6 (64) to get Pa
+    
+    // Convert 24-bit to signed int32
+    int32_t raw_temp = (int32_t)((bmp581_ut << 8) >> 8);
+    
+    // Temperature in 0.01°C (fixed point, 2 decimal places)
+    *temperature = (int32_t)(((int64_t)raw_temp * 100) / 65536);
```
pmattila: Does this need 64 bits? 
raw_temp is 24 bits. 100 is 7 bits.
That should fit into int32_t ?

pmattila (changes requested):
A few things to check

src/main/drivers/barometer/barometer_bmp581.c
pmattila: Also, if you are using bit shifts, you could as well use bit shifts here too.

src/main/drivers/barometer/barometer_bmp581.c
```diff
@@ -174,15 +174,11 @@
...
-typedef struct bmp581_calib_param_s {
-    // Reserved for future use if needed
-    uint8_t reserved;
-} __attribute__((packed)) bmp581_calib_param_t;
 
 // Driver state
 static uint8_t bmp581_chip_id = 0;
-static uint32_t bmp581_up = 0;  // Uncompensated pressure
-static uint32_t bmp581_ut = 0;  // Uncompensated temperature
+uint32_t bmp581_up = 0;  // Uncompensated pressure (accessible for unit tests)
```
pmattila: If only the unit test needs access, then it should be `STATIC_UNIT_TESTED`

NightHawk32: Right, good catch!

PR #361, opened 2026-01-02
MSP: Add single-index read support to MSP_MIXER_INPUTS to reduce over-air payload size

robthomson:
**Summary**
MSP_MIXER_INPUTS currently always returns the full mixer input table, which can result in large MSP payloads and timeouts when MSP is transported over low-bandwidth or high-latency links (for example over telemetry / RF links).

This change adds a backward-compatible single-index read mode to MSP_MIXER_INPUTS:

No request payload: existing behaviour, returns the full mixer input table

1-byte request payload (index): returns only the selected mixer input (rate, min, max)

This allows clients to read mixer inputs incrementally and significantly reduce MSP transfer size, while remaining fully compatible with existing tools.

**Details**

Implemented in mspFcProcessOutCommandWithArg()

Reuses existing MSP_MIXER_INPUTS command ID

Validates request payload size and index range

No changes to the legacy mspProcessOutCommand() path

**Request / Response formats**

_Legacy mode_
Request payload: none
Response payload: MIXER_INPUT_COUNT × (u16 rate, u16 min, u16 max)

_Indexed mode_
Request payload: u8 index
Response payload: u16 rate, u16 min, u16 max

**Motivation**
Over-air MSP links are sensitive to payload size. Large “read all” responses frequently cause retries or timeouts. Supporting indexed reads enables smaller MSP frames, faster UI response, and selective loading in configurators.

**Compatibility**

Fully backward compatible

Existing configurators continue to work unchanged

New clients may opt in to indexed reads

### DIFF
--- a/src/main/msp/msp.c
+++ b/src/main/msp/msp.c
@@ -2090,7 +2090,7 @@ static mspResult_e mspFcProcessOutCommandWithArg(mspDescriptor_t srcDesc, int16_
             serializeBoxReply(dst, page, &serializeBoxPermanentIdFn);
         }
         break;
-    case MSP_MIXER_INPUTS_INDEXED:
+    case MSP_GET_MIXER_INPUT:
         {
             const int rem = sbufBytesRemaining(src);
             if (rem != 1) {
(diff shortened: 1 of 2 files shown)


robthomson:
I have been developing against this for the past few days.

Zero issues.  It just works.

Issue #359, opened 2025-12-29
Send stats_total_flights, stats_total_time_s and stats_total_dist_m as telemetry data


heliflieger:
RF 4.6.0 introduces the transmission of flight count, seconds, and distance, as well as their display as MSP in the LUA app.
It would also be nice if these values were available via extended telemetry data. This would allow them to be easily displayed in a widget.

PR #370, opened 2026-01-10
#359 Add custom telemetry sensors for total flights, total time, and …

heliflieger:
Add STATS_TOTAL_FLIGHTS, STATS_TOTAL_TIME_S and STATS_TOTAL_DIST_M

### DIFF
--- a/src/main/telemetry/crsf.c
+++ b/src/main/telemetry/crsf.c
@@ -832,6 +832,11 @@ static telemetrySensor_t crsfCustomTelemetrySensors[] =
     TLM_SENSOR(DEBUG_5,                 0xDB05,   100,  3000,    0,     S32),
     TLM_SENSOR(DEBUG_6,                 0xDB06,   100,  3000,    0,     S32),
     TLM_SENSOR(DEBUG_7,                 0xDB07,   100,  3000,    0,     S32),
+
+    TLM_SENSOR(STATS_TOTAL_FLIGHTS,     0x1230,   200,  3000,    0,     U32),
+    TLM_SENSOR(STATS_TOTAL_TIME_S,      0x1231,   200,  3000,    0,     U32),
+    TLM_SENSOR(STATS_TOTAL_DIST_M,      0x1232,   200,  3000,    0,     U32),
+    
 };
 
 telemetrySensor_t * crsfGetNativeSensor(sensor_id_e id)
--- a/src/main/telemetry/sensors.c
+++ b/src/main/telemetry/sensors.c
@@ -29,6 +29,7 @@
 #include "pg/pg_ids.h"
 #include "pg/pilot.h"
 #include "pg/telemetry.h"
+#include "pg/stats.h"
 
 #include "build/debug.h"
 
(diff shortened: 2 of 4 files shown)


offer-shmuely:
I totally agree that it is a waste of bandwidth for something that barely changes.

However, in the current edgeTx implementation, there is rf2tlm.lua that run on the background, it can not share memory (only widgets can do that) 
for a widget to get that telemetry (as well as batt capacity, number of cells...) the widget will need to run msp calls, this poses two issues:
1. each widgets will need to have the msp code (complicated)
2. The widget will send msp on the same time the background script send it, so the result is not expected for both sides

This is why I copy all the rf2tlm to my dashboard, so there will not be conflicts with the rf2tlm, 
However users find it confusing, since the document say use it, and mine say don't 😭🤷‍♂️

There is an other method that the background script do, is to use area on the model configuration as it's storage, so it change the model name and global variables as he liked to, also i do understand the limitations, i don't like the script messing with my model 😥
We need to find an easier way.
I did asked from etx dev to add open key/value settings that will be available on the model for the use of scripts, but that still did not happened.

offer-shmuely:
I wanted to keep the widget data in edgeTx model with key/value api (like done in Jetti)
Not in the flight controller.

robthomson:
The global variable is, IMHO totally the wrong way, to do it.

It's pretty 'cool' and 'flexible'.  But it is essentially designed by a coder with the idea end users are coders.

This sort of feature needs to be better thought out as it's tailored too much around a bias to either edgetx or frsky, without delivering value to other telemetry systems (for example jeti and spektrum could easy have same sensors)

What's needed is a cleaner way of doing this thay works with end user in mind.

Conceptualy I love the idea, of pilot config and gvars.   It's a cleaver way of doing stuff.   But I don't think in this case it is the right place for flight timers etc.

heliflieger:
Maybe this is a better solution: https://github.com/rotorflight/rotorflight-lua-scripts/pull/125
Can you look into the PullRequest.
With this variant I read the statistics via MSP after disarming.

pmattila:
It appears these values are not updated during the flight, so it does not make much sense to send them over via telemetry.
We need to change the statistics first so that they are updated during the flight. And also add "statistics" for flight timer, and remaining time. Then it makes sense to send them over via telemetry.

heliflieger:
Ok, I understand.
Can I help with that?
Should we create a new ticket in which we define the necessary configuration parameters, telemetry keys, etc.?
I would not prefer a calculation in the transmitter. I would like to go with a single model in the transmitter and keep the transmitter dumb otherwise. As far as I can see, all relevant information is stored in Rotorflight.

robthomson:
> Ok, I understand. Can I help with that? Should we create a new ticket in which we define the necessary configuration parameters, telemetry keys, etc.? I would not prefer a calculation in the transmitter. I would like to go with a single model in the transmitter and keep the transmitter dumb otherwise. As far as I can see, all relevant information is stored in Rotorflight.

Are you using ethos or edge?

Ethos wise.. I can happily show you.

For edge.. you probably need to talk to @egonl as he heads up that side.  I have zero idea on the edge task architecture atm!

heliflieger:
I use EdgeTX. 
The language, whether LUA or C, is not the problem. However, I think we first have to adapt the rotorflight firmware. I can do that, but I would like to coordinate with someone first.
We can also rewrite Issue #359 and reuse it.

robthomson:
Firmware does not need to update.  It already does everything it needs.

Question is really can the existing background task be extended to detect this sort of event.

On ethos I basically do this:

https://github.com/robthomson/rotorflight-lua-ethos-suite/blob/2a30f347cb4fe26bf404f5479529ced9d37aaf64/src/rfsuite/tasks/sensors/msp.lua

PR #377, opened 2026-01-18
Spektrum smart telem

QuPsi:
Added Spektrum SMART telemetry functionality.

A new driver was created to handle the SRXL2 communication with the ESC. The "smart_esc.c" code was largely adapted from the srxl2.c file.

The Avian SMART ESC needs to be plugged into a UART port, but the power line needs to be pulled. A Y-connector splitting out the signal line to the UART port and the power lines to the servo bus needs to be created. This is necessary as the servo voltage (up to 8.4v) would overvolt the UART port.

wiring diagram

To activate, the new "SMART_ESC" feature needs to be activated and a serial port assigned (bitmask: 524288) with 115k baud.

When an Avian ESC and SRXL2 receiver is connected and configured, the bytes from the ESC are just passed directly to the SRXL2 receiver. 

The textgen menu from the ESC has been filtered out. Right now, only the throttle channel is fed to the ESC and for textgen to work, the 4 main channels need to be sent and this increases the channel data transmission time (and limits the throttle signal refresh rate). 

The ESC data is also injected into the escSensor data structure for blackbox logging (not tested yet). This requires the "ESC_SENSOR" feature to be activated. The configuration was modified to allow this feature to be active without assigning a serial port.

In the CLI, 'srxl2' brings up the status of the SRXL2 receiver communication. The 'smartesc' CLI command displays the current refresh rate and the telemetry request frequency.

The throttle refresh rate can be modified by 'smartesc rate <new rate>". Default is set to 250Hz. Can handle up to 600 Hz, but framing errors/collisions show up when looking at a logic analyzer above 250Hz. Not sure how it would effect governor performance at this time.

The telemetry request frequency can be modified by 'smartesc telem <number of frames>'. Default is set to 7. So telemetry data is requested every 7 frames. Lower numbers (with high refresh rates) can lead to framing errors/collisions.

Looking forward to hearing any feedback. Bear with me as I'm not a "coder" and AI did a lot of the heavy lifting.

## Summary by CodeRabbit

* **New Features**
  * Added SRXL2 ESC driver with telemetry data collection
  * Added SRXL2 ESC as new sensor protocol option
  * Added configurable throttle rate and telemetry intervals for SRXL2 ESC
  * Added CLI debugging commands for SRXL2 ESC


pmattila:
Build fails. Please fix.

PR #396, opened 2026-02-12
Extend MSP_MOTOR_CONFIG  to include a motor type definition

robthomson:
It would be usefull within RF to know the motor type in use on the heli.

This value can be used by the lua systems for:

- better audio alerts. 
For example in german you have 
   "Kraftstoff" and "Akku"
   
   On sites nitro; the other electric
   
- Themes
The value could be used to provide custom themes. For example we could present different displays depending on the motor type.  This would apply to Suite and RFC

I am not 100% sure this is the right place for the call - so putting it out there as a discussion point with a view to deciding if this is the right call for it or not.

## Release Notes

* **New Features**
  * Added motor type configuration option allowing users to select between electric and nitro motor modes via CLI and configuration interfaces. Default setting is electric.


breavyn:
`motorType` should be an enum, not just a bare integer.

robthomson:
> `motorType` should be an enum, not just a bare integer.

Done!

PR #421, opened 2026-03-07
Support for Spektrum SRXL2 ESC added.

QuPsi:
Spektrum Avian ESC SRXL2 bidirectional serial protocol added, with motor control and ESC sensor data. Select "SRXL2 ESC" in the drop down on the configuration menu. Under the motors tabs, enable pin swap if necessary. Telemetry must be enabled. Max update frequency set to 250Hz. Meant for helicopters using 1 SRXL2 ESC. Multiple ESCs not supported.

Can be used with any UART port, however, if not using the ports on the servo rail, power pin from ESC will need to be split and not feed into a UART.

Added the following CLI commands:
srxl2esc - status
srxl2esc telem - current telemetry request interval (default every 7 frames)
srxl2esc telem n - sets telemetry request interval to every n frames (recommend no lower than 5)

Telemetry data transmitted:
RPM
ESC amps
ESC volts
Throttle input percentage
Motor output percentage
FET temp
BEC voltage
BEC amp
BEC temp

Things to note:
1. Some Avian ESCs do not have RPM signal wires, therefore no RPM filtering and RF governor without a phase sensor.
2. Rotorflight governor requires fixed-wing mode and startup mode set to "normal" for proper governing.
3. No forward programming due to way it is implemented. Will clash with RF CMS.

## Summary by CodeRabbit

* **New Features**
  * Added full SRXL2 ESC support, including driver/task initialization, telemetry history/latest updates, and throttle/telemetry interval configuration.
  * Introduced SRXL2 as a selectable motor PWM protocol and added SRXL2 ESC as a selectable ESC-sensor protocol.
  * Added a `srxl2esc` CLI command to view driver status and set/query telemetry polling interval and throttle rate.
  * Added persistent SRXL2 ESC configuration parameters.
* **Documentation**
  * Updated serial port function documentation with the new `SRXL2_ESC` function constant.
* **Bug Fixes**
  * Improved ESC-sensor feature enabling logic to better account for SRXL2 ESC availability.


src/main/drivers/srxl2_esc.c
```diff
@@ -0,0 +1,1437 @@
...
+
+            /* If we got a valid handshake response, only exit when there is
+             * no pending or active TX left to flush. */
+            if ((state == Running || busMasterDeviceId != 0xFF) && writeBufferIdx == 0 && !srxl2escTxInProgress) {
+                break;
+            }
+
+            /* Small delay to prevent busy-loop CPU saturation */
+            delayMicroseconds(100);
+        }
```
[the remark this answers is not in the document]
QuPsi: The task scheduler comes online too late to respond to the ESC handshake. This ensures a handshake is sent out as soon as it's called to ensure the ESC handshake window doesn't close before the task scheduler comes online.

src/main/cli/cli.c
```diff
@@ -6534,6 +6540,73 @@ typedef struct {
...
 
+#ifdef USE_SRXL2_ESC
+static void cliSrxl2Esc(const char *cmdName, char *cmdline)
+{
+    extern void srxl2escSetThrottleRateHz(uint32_t rateHz);
+    extern uint32_t srxl2escGetThrottleRateHz(void);
+    extern void srxl2escSetTelemetryIntervalFrames(uint8_t frames);
+    extern uint8_t srxl2escGetTelemetryIntervalFrames(void);
+    extern bool srxl2escTelemetryRequested(void);
+    extern bool srxl2escDriverIsReady(void);
```
pmattila: This is poor coding style. In fact I have not seen this since the 80's.

Prototypes must be placed in a header file.

src/main/cli/cli.c
```diff
@@ -6534,6 +6540,73 @@ typedef struct {
...
+{
+    return "SRXL2 not present";
+}
+
+__attribute__((weak)) const char *srxl2_debug_lastErrorDM(void)
+{
+    return "None";
+}
+#endif
+
```
pmattila: This is not a typical way to do this in the codebase.
Please use the same coding style as rest of the code.

src/main/cli/settings.c
```diff
@@ -483,7 +483,7 @@ const char * const lookupTableErrorRelaxType[] = {
 
 #ifdef USE_ESC_SENSOR
 static const char * const lookupTableEscSensorProtocol[] = {
-    "OFF", "BLHELI32", "HOBBYWINGV4", "HOBBYWINGV5", "SCORPION", "KONTRONIK", "OMPHOBBY", "ZTW", "APD", "OPENYGE", "FLYROTOR", "GRAUPNER", "XDFLY", "RECORD",
+  "OFF", "BLHELI32", "HOBBYWINGV4", "HOBBYWINGV5", "SCORPION", "KONTRONIK", "OMPHOBBY", "ZTW", "APD", "OPENYGE", "FLYROTOR", "GRAUPNER", "XDFLY", "SRXL2", "RECORD",
 };
```
pmattila, line 489: Please don't randomly change whitespace.

src/main/config/feature.h
```diff
@@ -55,7 +55,7 @@ typedef enum {
     FEATURE_UNUSED_21              = BIT(21),
     FEATURE_UNUSED_22              = BIT(22),
     FEATURE_UNUSED_23              = BIT(23),
-    FEATURE_UNUSED_24              = BIT(24),
+    FEATURE_SRXL2_ESC              = BIT(24),
```
pmattila: Why is a new feature flag needed for this? Please explain the logic.

src/main/drivers/srxl2_esc.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
```
pmattila, line 18: File header missing.

src/main/drivers/srxl2_esc.h
```diff
@@ -0,0 +1,76 @@
...
+#ifdef USE_SRXL2_ESC
```
pmattila: This relies on the includer having the config defines available.
This may not be the case.

Please compare to other header files and include the requires system headers before the ifder

src/main/drivers/srxl2_esc.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include "common/time.h"
+
+#include "io/serial.h"
+
+#define SRXL2_ESC_PORT_BAUDRATE_DEFAULT    115200
+#define SRXL2_ESC_PORT_BAUDRATE_HIGH       400000
+#define SRXL2_ESC_DRIVER_TASK_FREQ_HZ      4000
+
+struct sbuf_s;
```
pmattila: But why?

src/main/drivers/srxl2_esc.h
```diff
@@ -0,0 +1,76 @@
...
+typedef struct srxl2esc_runtimeState_s {
+	uint16_t                    *channelData;      /* channel storage */
+	uint8_t                      channelCount;     /* number of channels */
+	uint16_t                     rxRefreshRate;    /* refresh period */
+	srxl2esc_readRawDataFnPtr  rcReadRawFn;      /* read raw channel helper */
+	srxl2esc_frameStatusFnPtr  rcFrameStatusFn;  /* frame-status callback */
+	srxl2esc_processFrameFnPtr rcProcessFrameFn; /* process-frame callback */
+	srxl2esc_getFrameTimeUsFn *rcFrameTimeUsFn;  /* timestamp helper */
+	timeUs_t                     lastRcFrameTimeUs;
+} srxl2esc_runtimeState_t;
```
pmattila, line 57: Please either align the member names, or don't use extra spaces at all.

src/main/drivers/srxl2_esc.h
```diff
@@ -0,0 +1,76 @@
...
+bool srxl2escGetLatestTelemetry(uint8_t *sensorId, uint8_t *secondaryId, uint8_t *dst, unsigned maxLen);
+bool srxl2escCopyLatestTelemetry(srxl2escTelemetrySnapshot_t *out, uint32_t *outSeq);
+
+void srxl2escDataReceive(uint16_t c, void *ctx);
+void srxl2escIdle(void);
+void srxl2escAttachPort(serialPort_t *p);
+
+bool srxl2escDriverInit(void);
+void srxl2escDriverTask(timeUs_t currentTimeUs);
+bool srxl2escDriverIsReady(void);
```
pmattila: All the prototypes should be extern in the header file

src/main/flight/motors.c
```diff
@@ -242,9 +242,14 @@ INIT_CODE void motorInit(void)
 {
```
pmattila, line 242: This function needs a re-think.

What should we do with the normal motor config and init in this case

src/main/sensors/esc_sensor.c
```diff
@@ -255,6 +265,120 @@ escSensorData_t * getEscSensorData(uint8_t motorNumber)
...
+        return;
+    }
+
+    srxl2escTelemetrySnapshot_t snap;
+    uint32_t seq = 0;
+    if (srxl2escCopyLatestTelemetry(&snap, &seq) && snap.valid) {
+        if (seq != srxl2escLastSeq) {
+            srxl2escLastSeq = seq;
+
+            if (getMotorCount() > 0 && featureIsEnabled(FEATURE_ESC_SENSOR)) {
```
pmattila, line 410: Why is this needed? is srxl2escDriverIsReady not considering this?

src/main/sensors/esc_sensor.c
```diff
@@ -255,6 +265,120 @@ escSensorData_t * getEscSensorData(uint8_t motorNumber)
...
+                int motorIndex = -1;
+                const int motorCount = getMotorCount();
```
pmattila: But why? motorCount was forced to 1 if SRXL2 ESC was initialised

pmattila (changes requested):
Many things still need attention. Back to draft.

src/main/cli/cli.c
QuPsi: Ok, I will fix, but I need more detail on what the issue is. Unfortunately, I'm not a coder and don't have much experience to know what's appropriate as far as style and best practices.

src/main/cli/cli.c
QuPsi: I've removed these lines. Not necessary.

src/main/cli/settings.c
QuPsi: Wasn't intentional. Fixed.

src/main/drivers/srxl2_esc.h
QuPsi: Why what exactly? Task frequency?

src/main/config/feature.h
QuPsi: I had to go back and look at the init sequence. I did this because the scheduler that handles the serial ports wakes up too late for the SRXL2 handshake (>250ms). By making it a feature, the srxl2escDriverInit function call happens early enough in the boot sequence to send out an initial handshake during the SRXL2 handshake window (the handshake finishes when the scheduler wakes up). 

If there's another way to call the srxl2escDriverInit function early in the boot sequence, please let me know and I'll modify.

src/main/sensors/esc_sensor.c
QuPsi: As it stands, no. I did not think it worthwhile to populate these values if ESC_SENSOR is not explicitly enabled. And I didn't want to create a condition where the driver doesn't work unless ESC_SENSOR is activated. Even without ESC_SENSOR fields being populated, the driver still needs to be active for SRXL2 communication to work.

src/main/sensors/esc_sensor.c
QuPsi: I'm hoping @AndersHoglund can chime in here. I haven't fully processed this part.

AndersHoglund: Motor count is forced to 1 at init otherwise there was issues in RFC motor page if not MOTOR had exactly 1 resource defined. If NONE or 2 (or more), the motor page would reflect that. I have no idea what is going on with motor count here in esc_sensor.

src/main/cli/cli.c
AndersHoglund: Hmm. Only external interface functions should be made visible in header files. Not internal functions.
If they are internal, functions can be moved in specific required order instead of using forward declaration prototypes. But that might not be possible if functions are cross referencing each other, this should however not happen as it would be poor and dangerous design.
You can of course have different kinds of header files, clearly named or in other ways made available as module internal vs external.

QuPsi:
Just giving a brief update. I have another issue to resolve with getting stale data. I haven't been able to touch this in a while, but I'm getting back to it. Hoping to have updates this week.

QuPsi:
Addressed as much as I could. Have not flown this commit yet and can't replicate the "stale" data issue yet. Will try to fly this weekend.

src/main/drivers/srxl2_esc.c
```diff
@@ -0,0 +1,1463 @@
...
+
+    srxl2escAttachPort(srxl2escDriverPort);
+
+    memset(&srxl2escDriverRuntime, 0, sizeof(srxl2escDriverRuntime));
+
+    /* Use a default local config here to avoid referencing escConfig()/rx types.
+     * If you want board-specific unit id or pin-swap, set them via a
+     * platform-specific setter or call `srxl2escInit` directly with values. */
+    srxl2esc_config_t localCfg = { .srxl2_unit_id = 0, .pinSwap = false };
+    if (!srxl2escInit(&localCfg)) {
```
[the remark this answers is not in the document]
QuPsi: srxl2escConfig_s has no srxl2_unit_id field

src/main/flight/motors.c
QuPsi: Why exactly does it need a rethink? Just so I can understand and possibly work on a solution.

AndersHoglund:
There are still dependencies to the rx subsystem, rx/rx.h and rx/srxl2_types.h are included. 
The two SRXL2 implementations, Rx (slave) and ESC (master) should have their shared definitions in a device independent common include file, general SRXL2 stack definitions. There must be no direct dependencies between specific devices, they should both depend on common general definitions. Or both have their own private defs, causing some duplication. 

There are also RX_FRAME_* constants used in this ESC device.  Makes no sense for an ESC having RX states. Or are those parts of the code some general SRXL2 stack code, handling both ESC and Rx and maybe other devices? A general SRXL2 stack should not be implemented in an ESC device driver. It should be in its own separate modules.

All in all, it looks as we would either need a bit more structured modularisation, or two isolated SRXl2 implementations (resulting in some duplications).

QuPsi:
> There are still dependencies to the rx subsystem, rx/rx.h and rx/srxl2_types.h are included. The two SRXL2 implementations, Rx (slave) and ESC (master) should have their shared definitions in a device independent common include file, general SRXL2 stack definitions. There must be no direct dependencies between specific devices, they should both depend on common general definitions. Or both have their own private defs, causing some duplication.
> 
> There are also RX_FRAME_* constants used in this ESC device. Makes no sense for an ESC having RX states. Or are those parts of the code some general SRXL2 stack code, handling both ESC and Rx and maybe other devices? A general SRXL2 stack should not be implemented in an ESC device driver. It should be in its own separate modules.
> 
> All in all, it looks as we would either need a bit more structured modularisation, or two isolated SRXl2 implementations (resulting in some duplications).

Hmm, I see your point. The references to rx.h is because I forked this code from the original srxl2 implementation. I think the easiest solution would be to have private defs but as you mentioned, this will lead to some duplication. The other route of having common definitions would require touching the rx.h file, which I'm very hesitant to do.

src/main/cli/cli.c
```diff
@@ -204,6 +205,12 @@ static bool commandBatchActive = false;
 static bool commandBatchError = false;
 #endif
 
+/* Forward declarations to avoid implicit-declaration when this helper is placed
+ * before the main CLI print function definitions. These match prototypes
+ * in cli.h. */
+extern void cliPrintLine(const char *str);
+extern void cliPrintLinef(const char *format, ...);
+
```
pmattila: I am not sure why these are needed here, if they are already in cli.h ?
If cli.h is not included, it certainly should be. That's the standard with all header files.

src/main/cli/cli.c
```diff
@@ -6694,6 +6701,52 @@ typedef struct {
...
+                    char *end = NULL;
+                    long interval = strtol(p, &end, 10);
+                    if (*end != '\0' || interval < 0 || interval > UINT8_MAX) {
+                        cliPrintErrorLinef(cmdName, "INTERVAL MUST BE BETWEEN 0 AND 255 FRAMES");
+                    } else {
+                        srxl2escSetTelemetryIntervalFrames((uint8_t)interval);
+                        if (interval == 0) {
+                            cliPrintLine("Telemetry polling disabled");
+                        } else {
+                            cliPrintLinef("Telemetry requested every %d frame(s)", (unsigned)interval);
```
pmattila: %d with unsigned argument

src/main/config/feature.h
pmattila: I am not sure how the driver init call is related to having a feature. 
Feature is just a flag that can be enabled or disabled. It should not dictate when the init is called?

src/main/drivers/srxl2_esc.c
```diff
@@ -0,0 +1,1469 @@
...
+ * Needs to be long enough to cover ESC reply start latency + line turnaround,
+ * otherwise FC TX can collide with the ESC reply on the half-duplex bus. */
+#define SRXL2_ESC_TELEMETRY_REQUEST_HOLDOFF_US  2500
+
+/* Upper bound for the request holdoff. Prevents long stalls (e.g. 10ms) when
+ * throttle update rate is configured low. */
+#define SRXL2_ESC_TELEMETRY_REQUEST_HOLDOFF_MAX_US  5000
+#define SRXL2_ESC_TELEM_HISTORY             15
+
+#define packet_HEADER                     0xA6
```
pmattila: all defines must be upper case

src/main/drivers/srxl2_esc.h
pmattila: `struct sbuf_s;`

looks random.

src/main/cli/cli.c
```diff
@@ -6694,6 +6701,52 @@ typedef struct {
...
+            cliShowParseError(cmdName);
+            return;
+        }
+    }
+
+    /* Print concise status: throttle refresh and telemetry rates */
+    {
+        const uint32_t throttleHz = srxl2escGetThrottleRateHz();
+        const uint8_t telemInterval = srxl2escGetTelemetryIntervalFrames();
+        const uint32_t telemHz = (telemInterval > 0) ? (throttleHz / (uint32_t)telemInterval) : 0;
```
pmattila: These are for printing, so there is no need for using a specific size integers.
`const unsinged int` would do just fine here, and would be faster as well, as the compiler does not need to emulate the size limitation.
It would then also print ok without the `(unsigned)` with printf

src/main/drivers/srxl2_esc.c
```diff
@@ -0,0 +1,1469 @@
...
+#include "drivers/time.h"
+#include "drivers/serial.h"
+#include "drivers/serial_uart.h"
+
+#include "io/serial.h"
+
+#include "drivers/srxl2_esc.h"
+#include "rx/srxl2_types.h"
+#include "rx/rx.h"
+#include "../sensors/esc_sensor.h"
```
pmattila: Should never need ../

src/main/target/common_pre.h
```diff
@@ -302,6 +302,7 @@ extern uint8_t _dmaram_end__;
 #if ((TARGET_FLASH_SIZE > 256) || (FEATURE_CUT_LEVEL < 10))
 #define USE_CAMERA_CONTROL
 #define USE_ESC_SENSOR
+#define USE_SRXL2_ESC
```
pmattila, line 306: We are on the limit with 512k flash, so this feature has to go to >512 category.
SORRY.

src/main/config/config.c
```diff
@@ -341,6 +349,25 @@ static void validateAndFixConfig(void)
     }
 #endif
 
+#if defined(USE_SRXL2_ESC)
+    if (findSerialPortConfig(FUNCTION_SRXL2_ESC)) {
+        if (!featureIsConfigured(FEATURE_SRXL2_ESC)) {
+            featureEnableImmediate(FEATURE_SRXL2_ESC);
```
pmattila: Features should not be enabled automatically.
They are designed to be master ON/OFF flags for the end-user to choose.

src/main/target/common_pre.h
QuPsi: Hmm, what are the implications of this? I just tried moving it to the 512 category and now it's non functional on the Nexus (unless I'm doing something else wrong).

src/main/config/feature.h
QuPsi: Making it a feature was only to get the init function to run early enough in the boot sequence. If the protocol is being used, that init _must_ execute within the first 250ms of power up.

I'll work on trying to remove the feature flag and report back.

QuPsi: Feature removed.

src/main/drivers/srxl2_esc.h
QuPsi: Removed.

AndersHoglund:
Still a lot of RX related stuff. Examples:

`#include "rx/srxl2_types.h"
`

And:

```
grep -ri rx_ src/main/drivers/srxl2_esc.*
src/main/drivers/srxl2_esc.c:#define SRXL2_ESC_TELEMETRY_RX_TIMEOUT_US   10000
src/main/drivers/srxl2_esc.c:    SRXL2_ESC_RX_FRAME_PENDING = 0,
src/main/drivers/srxl2_esc.c:    SRXL2_ESC_RX_FRAME_COMPLETE = (1 << 0),
src/main/drivers/srxl2_esc.c:    SRXL2_ESC_RX_FRAME_FAILSAFE = (1 << 1),
src/main/drivers/srxl2_esc.c:    SRXL2_ESC_RX_FRAME_PROCESSING_REQUIRED = (1 << 2),
src/main/drivers/srxl2_esc.c:    SRXL2_ESC_RX_FRAME_DROPPED = (1 << 3),
src/main/drivers/srxl2_esc.c:            srxl2escTelemetryRxDeadlineUs = now + SRXL2_ESC_TELEMETRY_RX_TIMEOUT_US;
src/main/drivers/srxl2_esc.c:        globalResult = SRXL2_ESC_RX_FRAME_DROPPED;
src/main/drivers/srxl2_esc.c:        globalResult = SRXL2_ESC_RX_FRAME_DROPPED;
src/main/drivers/srxl2_esc.c:    globalResult = SRXL2_ESC_RX_FRAME_DROPPED;
src/main/drivers/srxl2_esc.c:        globalResult = SRXL2_ESC_RX_FRAME_DROPPED;
src/main/drivers/srxl2_esc.c:    globalResult = SRXL2_ESC_RX_FRAME_PENDING;
src/main/drivers/srxl2_esc.c:            result |= SRXL2_ESC_RX_FRAME_PROCESSING_REQUIRED;
src/main/drivers/srxl2_esc.c:            result = (result & ~SRXL2_ESC_RX_FRAME_PENDING) | SRXL2_ESC_RX_FRAME_FAILSAFE;
src/main/drivers/srxl2_esc.c:            result = (result & ~SRXL2_ESC_RX_FRAME_PENDING) | SRXL2_ESC_RX_FRAME_FAILSAFE;
src/main/drivers/srxl2_esc.c:            result = (result & ~SRXL2_ESC_RX_FRAME_PENDING) | SRXL2_ESC_RX_FRAME_FAILSAFE;
src/main/drivers/srxl2_esc.c:        result |= SRXL2_ESC_RX_FRAME_PROCESSING_REQUIRED;
src/main/drivers/srxl2_esc.c:    if (!(result & (SRXL2_ESC_RX_FRAME_FAILSAFE | SRXL2_ESC_RX_FRAME_DROPPED))) {

```

For example FAILSAFE on a ESC interface? Does that make sense ??
/A

QuPsi:
I'll go through and fix the names. But for the srxl2_types, I'd really like not to duplicate this. Perhaps we should move it out of rx and have it as a general srxl2 reference that multiple devices can reference? 

I don't mind unique frame types for specific devices, I think that makes sense. But the srxl2_types would be general across different devices.

It's not a problem to remove the dependency but wanted to discuss before doing so.

AndersHoglund:
> I'll go through and fix the names. But for the srxl2_types, I'd really like not to duplicate this. Perhaps we should move it out of rx and have it as a general srxl2 reference that multiple devices can reference?
> 
> I don't mind unique frame types for specific devices, I think that makes sense. But the srxl2_types would be general across different devices.
> 
> It's not a problem to remove the dependency but wanted to discuss before doing so.

For common SRXL2 types there should be one common include file, located in a neutral place. Not sure where that would be, `src/main/common` could be one possible location. Absolutely not in the `src/main/rx` dir.  
Later, the Rx code could be refactored to use this common file.

And it is not just a naming issue, it is also functionality. The RX symbol names hint there are Rx functionality in here, in an ESC driver. Only changing the names will only hide that fact. Are those Rx related pieces not just dead code in an ESC driver? Suspect those are remainders from a general purpose multi device server, striped down for ESC only use. Not stripped hard enough.
For example, will there ever be a FAILSAFE situation with an ESC? Or a FRAME_DROPPED??
/A 

EDIT: Or do the RX_FRAME names refer to the serial port received data frames, from ESC? Not anything related to the Rx subsystem?

src/main/common/srxl2_types.h
```diff
@@ -0,0 +1,138 @@
...
+
+#define PACKED __attribute__((packed))
+
+typedef enum {
+    Disabled,
+    ListenForActivity,
+    SendHandshake,
+    ListenForHandshake,
+    Running
+} Srxl2State;
```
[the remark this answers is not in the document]
QuPsi: Won't be modifying legacy code.

src/main/common/srxl2_types.h
```diff
@@ -0,0 +1,138 @@
...
+    SetBindInfo = 0x5B,
+} Srxl2BindRequest;
+
+typedef enum {
+    NotBound = 0x0,
+    DSM2_1024_22ms = 0x01,
+    DSM2_1024_MC24 = 0x02,
+    DMS2_2048_11ms = 0x12,
+    DMSX_22ms = 0xA2,
+    DMSX_11ms = 0xB2,
```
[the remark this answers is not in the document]
QuPsi: Won't be modifying legacy code.

src/main/common/srxl2_types.h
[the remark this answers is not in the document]
QuPsi: Not at this time.

src/main/common/srxl2_types.h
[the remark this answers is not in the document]
QuPsi: Not at this time.

pmattila (approved):

PR #434, opened 2026-03-27
Set default for pin swap to be on 

robthomson:
Anybody using FBUS master will likely be on nexus or 007

The feature always requires pinswap to be on - so setting as the default value makes sense.

## Summary by CodeRabbit

* **Configuration Changes**
  * FBUS Master serial port TX/RX pin swapping is now enabled by default.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -241,7 +241,7 @@ the actual values are calculated automatically (#332).
 
 `fbus_master_frame_rate` defaults to 500 Hz.
 
-`fbus_master_pinswap` defaults to OFF (0).
+`fbus_master_pinswap` defaults to ON (1).
 
 `fbus_master_inverted` defaults to ON (SERIAL_INVERTED), which is the standard for FBUS receivers.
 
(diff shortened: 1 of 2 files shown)


AndersHoglund:
Not true. Pinswap on or off depends on what UART port and pin is used. You can NOT assume all users will connect to a UART Rx-pin. But OK, this is only setting the default. Still changeable by the user I assume, not a forced setting.
 
You can however force bi-directional as that is required on any UART. Or just set as a default too.
Forcing a setting should be done using the  "...ConfigMutable()->..."-way in a validateAndFix() init function.

In the SRXL2_ESC PR bidirectional is forced but pinswap is not touched. Same with more serial devices.
Would be nice if we have a common way of handling things like this. 
/A

robthomson:
> Not true. Pinswap on or off depends on what UART port and pin is used. You can NOT assume all users will connect to a UART Rx-pin. But OK, this is only setting the default. Still changeable by the user I assume, not a forced setting.
> 
> You can however force bi-directional as that is required on any UART. Or just set as a default too. Forcing a setting should be done using the "...ConfigMutable()->..."-way in a validateAndFix() init function.
> 
> In the SRXL2_ESC PR bidirectional is forced but pinswap is not touched. Same with more serial devices. Would be nice if we have a common way of handling things like this. /A

Fair.  Could also be done that way.

pmattila:
There is no point in assuming pinswap when no FC has a port marked for "FBUS out" or even "FBUS".
It's hard to say where people would be plugging it in. Could be SBUS could be AUX could be something else.
Both ways, the default could be wrong.

PR #438, opened 2026-03-30
Refinement of the FBUS sensors Implementation

NightHawk32:
[commit_review_a8c3abcb.pdf](https://github.com/user-attachments/files/26351082/commit_review_a8c3abcb.pdf)
Some small adjustments based on the findings in this document. Item 1 is nonsense and can be disregarded.

## What changed
- Fixed dead scan logic in `fbusMasterTakeNextScanPhysId()` (src/main/drivers/fbus_master.c:109) by removing the unreachable `FC_COMMON_ID` (src/main/drivers/fbus_master.c:64) skip branch.
- Removed the unreachable `ESC_SENSOR_PROTO_FBUS` (src/main/sensors/esc_sensor.c:4214) switch case from `escSensorInit()` (src/main/sensors/esc_sensor.c:4202).
- Restored bus-servo diff/dump behavior in `printServo()` (src/main/cli/cli.c:1999) so changed bus-servo config is still emitted even when no FBUS/SBUS-out port is currently active, and so both copied and current serial configs are considered.
- Added non-destructive forwarded-frame access with `fbusSensorPeekForwardedFrame()` (src/main/drivers/fbus_sensor.c:369) and updated `getFbusSensorValue()` (src/main/telemetry/sensors.c:131) to peek instead of consume, avoiding competition with the actual FBUS uplink forwarding path.
- Removed the unused `FBUS_GPS_LONGITUDE_BASE` (src/main/drivers/fbus_sensor.h:58) constant and clarified that latitude/longitude share the same app ID range and are distinguished by bit 31, matching the FrSky S.Port spec in frsky_documents/Smart_Port_Protocol_V3.5.8.md.
- Deduplicated `fbusGetBit()` (src/main/rx/fbus.h:23) by moving it into fbus.h and updating src/main/drivers/fbus_master.c:45 to include that header.

ToDo: Testing

- [x] servos CLI tool
- [x] sensors forwarding

Cross checked with @robthomson  SPORT implementation and the previous FBUS PR. Essentially, no fundamental changes, just some code cleanup as discussed.


robthomson:
I have tested this one with lukas this afternoon to confirm no regressions.

pmattila (approved):

PR #441, opened 2026-04-03
Added support for AM32 and BLHeli_S programming on F411

egonl:
Tested on a DarwinFPV 15A running Bluejay.


robthomson:
@egonl I believe this PR is no longer valid as it is pat of the other fix we did?

This one

https://github.com/rotorflight/rotorflight-firmware/pull/440

egonl:
You're right Rob.

PR #443, opened 2026-04-11
Add msp command to retrieve the current setpoint for each axis

breavyn:
To be used for accurately displaying rates preview in the configurator.

## Summary by CodeRabbit

* **New Features**
  * Added an MSP command to retrieve the current setpoint values (returns four axis setpoints).

* **Documentation**
  * Changelog/documentation updated with the new MSP command entry.

### DIFF
--- a/src/main/msp/msp.c
+++ b/src/main/msp/msp.c
@@ -1374,7 +1374,7 @@ static bool mspProcessOutCommand(int16_t cmdMSP, sbuf_t *dst)
 
     case MSP_SETPOINT:
         for (int i = 0; i < 4; i++) {
-            sbufWriteU16(dst, lrintf(getSetpoint(i)));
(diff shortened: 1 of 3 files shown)


src/main/msp/msp.c
```diff
@@ -1372,6 +1372,12 @@ static bool mspProcessOutCommand(int16_t cmdMSP, sbuf_t *dst)
         }
         break;
 
+    case MSP_SETPOINT:
+        for (int i = 0; i < 4; i++) {
+            sbufWriteS16(dst, lrintf(getSetpoint(i)));
```
pmattila: The setpoint is in range -1000...1000 at extreme maximum.
Probably could use the 16 bits wisely and do *10 here.

breavyn: added x10 multiplier

pmattila (approved):

PR #445, opened 2026-04-16
Fix buffer overflow in RX protocols

pmattila:
## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened receiver validation to reject malformed frames, enforce payload/length bounds, prevent buffer overruns, and avoid off-by-one channel writes across multiple protocols.
  * Improved handling of unexpected or out-of-range channel/count values to avoid invalid unpacking.
  * Hardened cross-context state visibility to ensure reliable real-time receiver behavior.

* **Tests**
  * Updated unit test expectations to reflect stricter frame validation.


pmattila:
@coderabbitai full review please

PR #448, opened 2026-04-19
Moved SPORT to separate task

NightHawk32:
This is just to remove the dependency between the telemetry feature and SPORT Master for the sake of being easy to understand.
The separate task makes it easy to get SPORT update rates and reduces the risk of running into issues because of the SPORT feature being woven into the telemetry task.
E.g., using SPORT GPS for RTH without having a telemetry-capable RC link.

built on top of https://github.com/rotorflight/rotorflight-firmware/pull/438


robthomson:
Tested and all works as expected

pmattila (approved):

PR #452, opened 2026-04-26
Add freq_input_minhz parameter

pmattila:
Currently the minimum frequency for the RPM input is 25Hz. This is fine with electric motors / ESCs, but not low enough with magnetic sensors on the main shaft/gear. 
This change allows changing the min frequency to suite. Probably 2Hz is suitable for magnet on main shaft sensing.

## Summary by CodeRabbit

* **New Features**
  * Frequency sensors now support configurable per-input minimum and maximum thresholds; inputs below the minimum are treated as zero and updates are gated by the maximum.
  * Added a CLI parameter to set the minimum input frequency (valid range 1–100 Hz) and sensible defaults for min/max.

* **Documentation**
  * Updated docs to describe the new frequency configuration parameter and defaults.

### DIFF
--- a/Changes.md
+++ b/Changes.md
@@ -212,6 +212,8 @@ the actual values are calculated automatically (#332).
 
 `fbus_master_inverted` parameter added (ON/OFF). Controls electrical inversion of the FBUS Master UART output.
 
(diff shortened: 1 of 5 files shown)


pmattila:
@coderabbitai full review

PR #467, opened 2026-06-08
Parse the input throttle value of FLYROTOR ESC

ZyonMo:
Change the unused Mode data of FLYROTOR ESC to the input throttle value so that the FC black box can record the input throttle value data.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Fly telemetry frame interpretation to align with the actual field layout.
  * ESC telemetry now correctly extracts and reports throttle information (alongside existing power/PWM and temperature readings).


ZyonMo:
Test complete, data is normal.

pmattila (approved):

PR #478, opened 2026-07-20
 Allow old and new mode for XDFLY/ZTW/OMPHOBBY ESC telemetry

robthomson:
**Summary**

Older XDFLY/ZTW/OMPHOBBY ESC firmware only broadcasts unsolicited telemetry frames and just needs a plain RX wire. Newer firmware additionally supports a bidirectional handshake for parameter read/write.

Previously halfDuplex was force-enabled for these three protocols in validateAndFixEscSensorConfig(), regardless of the user's setting, which required half-duplex-capable wiring even for old firmware that never uses it.

halfDuplex is now left under user control (esc_sensor_halfduplex) for XDFLY/ZTW/OMPHOBBY (still forced for GRAUPNER). xdflySensorInit() takes a bidirectional flag: when true, it behaves as before (handshake + parameter read/write); when false, it just passively decodes telemetry frames with no TX/handshake state machine.

**Testing**
Fix has been tested on old and new (un unpublished) ztw firmwares.   
Fix has been checked for good measure on xdfly firmware.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bidirectional ESC sensor initialization and request handling.
  * Updated protocol processing for improved compatibility with supported ESC telemetry protocols.
  * Corrected half-duplex configuration behavior for Graupner, XDFLY, OMPHobby, and ZTW protocols.
  * Improved parameter commit handling during sensor initialization.

### DIFF
--- a/src/main/sensors/esc_sensor.c
+++ b/src/main/sensors/esc_sensor.c
@@ -4046,12 +4046,26 @@ static int8_t xdfly_accept(uint16_t c)
     return 0;
 }
 
-static serialReceiveCallbackPtr xdflySensorInit(uint8_t sig)
+static serialReceiveCallbackPtr xdflySensorInit(uint8_t sig, bool bidirectional)
 {
-    rrfsmCrank  = xdfly_crank_unc_setup;
+    rrfsmStart = NULL;
+    rrfsmCrank = bidirectional ? xdfly_crank_unc_setup : NULL;
     rrfsmDecode = xdfly_decode;
     rrfsmAccept = xdfly_accept;
-    paramCommit = xdfly_param_commit;
+    rrfsmBootDelayMs = 0;
+    rrfsmMinFrameLength = bidirectional ? 0 : XDFLY_HEADER_LENGTH;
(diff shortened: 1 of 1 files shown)


src/main/sensors/esc_sensor.c
```diff
@@ -4292,11 +4292,6 @@ void INIT_CODE validateAndFixEscSensorConfig(void)
 {
     switch (escSensorConfig()->protocol) {
         case ESC_SENSOR_PROTO_GRAUPNER:
-        case ESC_SENSOR_PROTO_XDFLY:
-        case ESC_SENSOR_PROTO_OMPHOBBY:
-        case ESC_SENSOR_PROTO_ZTW:     
-            escSensorConfigMutable()->halfDuplex = true;
-            break;
```
breavyn, line 4299: This changes the behaviour of GRAUPNER.

robthomson: Yes.. should have had this a draft - not finished testing the work yet!

pmattila (approved):

PR #491, opened 2026-08-29
Free up flash memory: CMS remaoval

NightHawk32:
Removes the on-device CMS menu and its telemetry-integrated display backends (SRXL/CRSF/HoTT), along with associated menu hooks and CLI settings, to reclaim flash/RAM. Configuration remains available via CLI and Configurator; nothing else changes.

Savings on STM32F7X2: flash usage down 4.7% (462,060 → 440,344 B, ~21.2 KB freed, 94.0% → 89.6% used), text size down 4.3%, RAM (data+bss) down ~1.3%.

This shows what could be saved. Decision pending.
Configurator nees to reflect this change too ...

## Summary by CodeRabbit

* **Removed Features**
  * Removed the on-screen configuration menu system and its associated flight-controller menus.
  * Removed display-based menu support through CRSF, HoTT, and Spektrum telemetry links.
  * Removed CMS-specific configuration options, status messages, and menu controls.
  * Camera key simulation and related controls now operate without menu-state restrictions.

* **Build and Configuration**
  * CMS is no longer included in firmware builds or available as a selectable feature.
  * OSD configuration no longer includes the CMS background setting.


AndersHoglund:
Wow, this is a brutal way to shrink code size. Too brutal, when all you need to do is to add one target definition line (or undef really):

```
andho@pucko:~/projects/rotorflight/rotorflight-firmware$ git diff
diff --git a/src/main/target/common_post.h b/src/main/target/common_post.h
index 6a16fd857..2cb66a280 100644
--- a/src/main/target/common_post.h
+++ b/src/main/target/common_post.h
@@ -24,6 +24,8 @@

 #include "build/version.h"

+#undef USE_CMS
+
 #if defined(USE_VTX_RTC6705_SOFTSPI)
 #define USE_VTX_RTC6705
 #endif

```
No need to wipe out all source code for it. There might be users out there that still want CMS in their FCs. 
We either keep it as is, or at least keep it for possible custom builds. 

Cloud based firmware builds with options to add or remove features would be the prefered way to save Flash space.
/A

NightHawk32:
Okay, agreed, so then I will just disable the feature.
Regarding the cloud based build process: We need people to tackle this and also resources to maintain it. Not as easy as it sounds.

